PR #5, opened 2017-01-26
Build Helm Repos (packages & indexes)

goruha:
## what
* Move `Makefile` targets from `cloudposse/charts` to this repo
* Generalize targets to work on arbitrary repo paths

## why
* To make logic more re-usable: https://github.com/cloudposse/charts/pull/42 which introduces the concept of a helm chart server

## who
@cloudposse/engineering


osterman (approved):

modules/helm/Makefile.repo
```diff
@@ -0,0 +1,101 @@
+REPO_URL?=https://charts.cloudposse.com
+
+ifneq ($(TRAVIS_PULL_REQUEST_BRANCH),)
```
osterman, line 3: What do you think about moving this to `modules/helm/Makefile.travis`?

goruha: Until we will have point where we can reuse this code I prefer to leave it in one place.

PR #7, opened 2017-01-31
Change Helm Directory Structure

goruha:
## what
* make it more closely resemble http://github.com/kubernetes/charts

## why
* easier to deploy to object stores like S3 and to bundle in docker images (see: https://github.com/cloudposse/charts/pull/42)

## who
@cloudposse/engineering

### DIFF
--- a/modules/helm/Makefile.repo
+++ b/modules/helm/Makefile.repo
@@ -69,6 +69,7 @@ helm\:repo\:package:
     	xargs -n 1 -I '{}' sh -c 'echo && echo "# Updating dependencies for {}" && $(HELM) dependency build --debug {}'
 	@find ./$(REPO_PATH) -maxdepth 1 -mindepth 1 -type d '!' -name packages  | \
     	xargs -n 1 -I '{}' sh -c 'echo && echo "# Building package for {}" && $(HELM) package --debug {}'
+   	@mkdir -p ./packages/$(REPO_PATH)
 	@find . -maxdepth 1 -mindepth 1 -type f -name '*.tgz' | \
 		xargs -I '{}' mv {} ./packages/$(REPO_PATH)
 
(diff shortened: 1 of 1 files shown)


modules/helm/Makefile.repo
```diff
@@ -59,25 +59,26 @@ helm\:repo\:fix-perms:
...
 
 .PHONY : helm\:repo\:lint
-## Lint library charts
+## Lint charts
 helm\:repo\:lint:
 	$(call assert-set,REPO_PATH)
-	@echo "## Linting charts $(shell pwd)/library"
-	@find ./$(REPO_PATH)/library -maxdepth 1 -mindepth 1 -type d | \
+	@echo "## Linting charts $(shell pwd)"
+	@find ./$(REPO_PATH) -maxdepth 1 -mindepth 1 -type d | \
```
osterman, line 81: I don't think this should be prefixed with `./`; that is what `REPO_PATH` is for

modules/helm/Makefile.repo
```diff
@@ -86,8 +87,8 @@ helm\:repo\:lint:
 helm\:repo\:clean:
 	$(call assert-set,REPO_PATH)
 	@rm -f *.tgz
-	@find ./$(REPO_PATH)/library -maxdepth 3 -mindepth 3 -type f -name '*.tgz' -delete
-	@rm -f ./$(REPO_PATH)/packages/index.yaml ./packages/*.tgz
+	@find ./$(REPO_PATH) -maxdepth 3 -mindepth 3 -type f -name '*.tgz' -delete
+	@rm -f ./packages/$(REPO_PATH)/index.yaml ./packages/*.tgz
```
osterman, line 91: We need to define 3 new ENVs
`REPO_NAME ?= incubator # for example`
`REPO_PATH ?= ./$(REPO_NAME)`
`PACKAGE_PATH ?= ./packages/$(REPO_NAME)`

modules/helm/Makefile.repo
goruha: Done

modules/helm/Makefile.repo
goruha: Done

PR #15, opened 2017-02-18
Add Cross Platform Support to Helm

osterman:
## what
* Install `helm` based on OS
* Remote `local` repo by default
* Install to `/usr/local/bin`

## why
* Make it compatible with OSX
* `/usr/bin/` is read-only on OS X
* `local` repo is not used by cloudposse

## who
@cloudposse/engineering


goruha:
ok

PR #24, opened 2017-03-21
Switch to getglide.sh

osterman:
## what
* Swap out `glide.sh` for `getglide.sh`
* Test that `go:deps-build` and `go:deps-dev` work
* Use `sudo: true` as a *hack* to avoid container based infrastructure on travis which leads to [rate limit](https://github.com/Masterminds/glide/issues?utf8=✓&q=is%3Aissue%20is%3Aopen%20rate%20limit) issues
* Use verbose output (`sh -x`) 

## why
* `glide.sh` expired and is pending deleted; problem with registrar getting it renewed
* https://github.com/Masterminds/glide/issues/784
* Verbose output enabled because when installing glide the script does not output accurate error messages when encountering GitHub rate limiting errors
* Using VMs instead of containers reduces likelihood of hitting GitHub rate limits because more containers share IPs that VMs.

## who
@goruha

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 language: go
 go:
   - 1.7.x
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,18 @@
+language: go
+go:
+  - 1.7.x
+addons:
+  apt:
+    packages:
+    - git
+    - make
+    - curl
+
+install:
+- make init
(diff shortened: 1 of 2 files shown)


osterman:
@goruha do you want to switch to this method? https://github.com/minimum2scp/geco/pull/21/files

goruha (approved):

PR #28, opened 2017-04-14
Fix helm indexing

goruha:
## What
* Added merge option for helm repo index

## Why
* You need old index be merged with new index to preserve info about previous versions of chart


osterman (approved):
Great fix!

PR #33, opened 2017-07-25
allow docker commands only for trusted builds in Travis

cicdteam:
# What

- added workflow to define Travis builds are trusted or not
- added conditions to target `travis:docker-tag-and-push` for trusted builds only
- added target `travis:docker-login` with conditions for trusted builds only
- added url shortner for `build-harness` and README updated
- template for `templates/docker.travis.yml` updated

# Why

- avoid run `docker login` and `docker push` commands for un-trusted builds in Travis (PR's from forks)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ It's 100% Open Source and licensed under [APACHE2](LICENSE).
 At the top of your `Makefile` add, the following...
 
 ```make
--include $(shell curl -so .build-harness "https://raw.githubusercontent.com/cloudposse/build-harness/master/templates/Makefile.build-harness"; echo .build-harness)
+-include $(shell curl -sSL -o .build-harness "https://git.io/build-harness"; echo .build-harness)
 ```
 
 This will download a `Makefile` called `.build-harness` and include it at run-time. We recommend adding the `.build-harness` file to your `.gitignore`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ It's 100% Open Source and licensed under [APACHE2](LICENSE).
 At the top of your `Makefile` add, the following...
 
 ```make
--include $(shell curl -so .build-harness "https://raw.githubusercontent.com/cloudposse/build-harness/master/templates/Makefile.build-harness"; echo .build-harness)
+-include $(shell curl -sSL -o .build-harness "https://git.io/build-harness"; echo .build-harness)
```
goruha, line 15: Cool

modules/travis/Makefile.docker
```diff
@@ -1,5 +1,32 @@
...
+endif
+
+## Login into docker hub
+travis\:docker-login:
+ifeq ($(ALLOW_DOCKER_ACTIONS),true)
+	@make docker:login
+else
+	@echo "Skipping docker login for untrusted builds"
+endif
+
```
goruha, line 24: Did you try this if\else outside of travis env?

goruha: ok. I see

goruha (approved):

README.md
osterman: Ya, let's do it. My only concern is we don't know who runs `git.io`

osterman (approved):
Great fix!

PR #38, opened 2017-12-18
Feature use new helm

goruha:
## What
* Use new helm version
* Added additional helm repos

## Why
* Follow helm update
* We need that charts as dependencies


modules/helm/Makefile.repo
```diff
@@ -43,6 +43,8 @@ helm\:repo\:add-current:
 	
 helm\:repo\:add:
 	@$(HELM) repo add kubernetes-charts http://storage.googleapis.com/kubernetes-charts
+	@$(HELM) repo add opsgoodness http://charts.opsgoodness.com
+	@$(HELM) repo add inclubator https://kubernetes-charts-incubator.storage.googleapis.com
```
const-bon: Typo `inclubator` should be `incubator`. But I suggest using official name `kubernetes-charts-incubator`

const-bon (approved):

PR #44, opened 2018-02-24
Switch to Slashes in Make Targets

osterman:
## what
* Switch to using `/` in target names
* Support `:` in target names for backward compatibility
* Fix `readme/build` on OSX

## why
* `:` is not truly supported in make targets as they do not work when used as target dependencies
* `grep -P` not supported on BSD `grep` (OSX)

## example

this does not work:

```
test\:deps:
  echo "Hey"

test\:me: test:deps
```

This will not run `test:deps` and it silently skips it. Escaping the `:` in the dependency causes make error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ addons:
     - curl
 
 install:
-- make go:deps-build
-- make go:deps-dev
+- make go/deps-build
+- make go/deps-dev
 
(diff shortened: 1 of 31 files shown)


aknysh (approved):
Thanks for fixing `doc_generate.sh` on OSX

PR #49, opened 2018-03-29
Fix `make readme`

aknysh:
## what
* Fix `readme/init` target
* `template/build` target explicitly depends on `template/deps`  target
* Fix typos

## why
* Wrong output file name in `readme/init`
* Nobody knows that they need to call `template/deps` to install `gomplate` when calling `make readme` (`gomplate` is ignored in `vendor/.gitignore`)


osterman (approved):

osterman:
Thanks @aknysh !

PR #55, opened 2018-05-02
Fix path

osterman:
## what
* Do not use `which`, especially with `:=`
* Set search path to include vendor directory

## why
* If binary is not already installed, `which` will fail
* Search path did not look for vendor directory
* Using `:=` prevents reassignment


osterman:
Related: https://travis-ci.org/cloudposse/charts/builds/373774418

goruha (approved):

Issue #63, opened 2018-06-18
`helm/toolbox/upsert`: ERROR: Unable to connect to helm server 


osterman:
## what
* we’re seeing this error intermittently from the ‘Deploy backing services with helmfile codefresh’ step:
```
ERROR: Unable to connect to helm server                                                                         
make: *** [//build-harness/modules/helm/Makefile:24: helm/toolbox/upsert] Error 1
```

## why
* It could be simply related to timeouts.

osterman:
@alebabai any update on this?

PR #69, opened 2018-06-22
helm-toolbox: improve upsert code

alebabai:
## what
* add replace wait function by OTB helm --wait flag
* remove failure on empty tiller version
* add retries

## why
* #63

### DIFF
--- a/bin/helm_toolbox.sh
+++ b/bin/helm_toolbox.sh
@@ -3,6 +3,8 @@
 export TIMEOUT=3
 export STDOUT=${STDOUT:-/dev/null}
 
+RETRIES=5
+
 # helper functions
 function info() { 
 	echo -e "\e[32mINFO:\e[0m $1"; 
@@ -17,19 +19,6 @@ function check() {
 	command -v "$1" >/dev/null 2>&1 || err "$1 not installed!"; 
 }
 
-# wait for helm to become operational
-wait_for_helm() {
-	info "Waiting for helm tiller..."
-	while true; do
-		status=$(echo $(kubectl get pods -l app=helm -l name=tiller --show-all=false -o=custom-columns=STATUS:.status.phase --no-headers=true -nkube-system))
-		info "Helm status: $status"
-		if [ "$status" = "Running" ]; then
-			break;
-		fi
-		sleep $TIMEOUT
-	done
-}
-
 function set_context() {
   if [ -n "$KUBE_CONTEXT" ]; then
 	info "Using ${KUBE_CONTEXT} kube context"
@@ -37,20 +26,25 @@ function set_context() {
   fi
 }
 
+function retry() {
+  attempt=1
+  until [[ $n -ge $2 ]]
+  do
+	info "Perform attempt - $attempt"
+    $1 && break
+    n=$[$attempt]
+    sleep 15
+  done
+}
+
 function upsert() {
 	helm_version=$(helm version --client --short | grep -Eo "v[0-9]\.[0-9]\.[0-9]")
 	tiller_version=$(timeout $TIMEOUT helm version --server --short | grep -Eo "v[0-9]\.[0-9]\.[0-9]")
 
-    if [ -z "$tiller_version" ]; then
-      err "Unable to connect to helm server"
-    fi
-
 	if [ "$helm_version" != "$tiller_version" ]; then
-		info "Helm version: $helm_version, differs with tiller version: $tiller_version"
+		info "Helm version: $helm_version, differs with tiller version: ${tiller_version:-'not installed'}"
 		info "Upgrarding tiller to $helm_version"
-		helm init --upgrade --force-upgrade > $STDOUT
-		wait_for_helm
-		sleep 3
+		helm init --upgrade --force-upgrade --wait > $STDOUT
 		info "Helm version"
 		helm version --short | sed 's/^/  - /'
 	else
@@ -65,7 +59,7 @@ if [ "$1" == "upsert" ]; then
 	check helm
 	check timeout
 	set_context
-	upsert
+	retry upsert $RETRIES
 else
 	err "Unknown commmand"
 fi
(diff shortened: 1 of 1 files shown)


bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
 function set_context() {
   if [ -n "$KUBE_CONTEXT" ]; then
 	info "Using ${KUBE_CONTEXT} kube context"
 	kubectl config use-context ${KUBE_CONTEXT}
   fi
 }
 
+function retry() {
+  attempt=1
+  until [[ $n -ge $2 ]]
```
sarkis: Should this be `until [[ $attempt -ge $2 ]]`? Where is `$n` set?

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
+  do
+	info "Perform attempt - $attempt"
+    $1 && break
+    n=$[$attempt]
```
sarkis: Related to previous comment, I don't think we want `$n` here but instead something like: `attempt=$[$attempt+1]`

sarkis (changes requested):
inline comments

bin/helm_toolbox.sh
alebabai: sorry, mistyped.
fixed now.

bin/helm_toolbox.sh
```diff
@@ -3,6 +3,8 @@
 export TIMEOUT=3
 export STDOUT=${STDOUT:-/dev/null}
 
+RETRIES=5
```
osterman: Make this overridable by env

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
-
 function set_context() {
   if [ -n "$KUBE_CONTEXT" ]; then
 	info "Using ${KUBE_CONTEXT} kube context"
 	kubectl config use-context ${KUBE_CONTEXT}
   fi
 }
 
+function retry() {
+  attempt=0
```
osterman: Use local

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
+  until [[ $attempt -ge $2 ]]
```
osterman: Name $2

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
+  do
+	info "Perform attempt - $attempt"
+    $1 && break
```
osterman: Name $1

osterman (changes requested):

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
 	info "Using ${KUBE_CONTEXT} kube context"
 	kubectl config use-context ${KUBE_CONTEXT}
   fi
 }
 
+function retry() {
+  attempt=0
+  until [[ $attempt -ge $2 ]]
+  do
+	info "Perform attempt - $attempt"
```
sarkis: 2 things here:

1. indentation not necessary (format fix)

2. I suggest changing this to: `info "Perform attempt - $[attempt+1]"` otherwise the first attempt will log as "Perform attempt - 0"

sarkis (changes requested):
minor changes requested

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,32 @@ function check() {
...
-}
-
 function set_context() {
   if [ -n "$KUBE_CONTEXT" ]; then
 	info "Using ${KUBE_CONTEXT} kube context"
 	kubectl config use-context ${KUBE_CONTEXT}
   fi
 }
 
+function retry() {
```
osterman: `local limit=$2`

osterman: Also, I would recommend putting arguments that affect `retry` first, so that `$*` can be the ones for the command being retried. e.g. think about how `chamber` and `aws-vault` work.

bin/helm_toolbox.sh
osterman: `RETRIES=${RETRIES:-5}`

bin/helm_toolbox.sh
```diff
@@ -65,7 +59,7 @@ if [ "$1" == "upsert" ]; then
 	check helm
 	check timeout
 	set_context
-	upsert
+	retry upsert $RETRIES
```
osterman: suggest using `retry $RETRIES upsert` instead. That way, if we ever add arguments to `upsert`, they can follow.

osterman (approved):

alebabai:
@sarkis could you look into these changes, please?

bin/helm_toolbox.sh
```diff
@@ -17,40 +19,33 @@ function check() {
...
   fi
 }
 
+function retry() {
+  local attempt=0
+	local limit=$1
+	local command=${@:2}
+  until [[ $attempt -ge $limit ]]
+  do
+		info "Perform attempt - $[$attempt+1]"
```
sarkis: Indent looks off on this line and the one below. What could help with this in the future is adding a `*.sh` config here: https://github.com/cloudposse/build-harness/blob/master/.editorconfig and ensuring you have the `editorconfig` plugin installed for your editor.

Here is an example of sh editorconfig we use https://github.com/cloudposse/terraform-root-modules/blob/master/.editorconfig#L22-L24

sarkis (approved):
Looks good - minor nitpick inline. I'd say OK to merge after fixing the indent issues and update editorconfig / setup plugin at a later time in a different PR

bin/helm_toolbox.sh
osterman: Yea, make sure your editor supports `.editorconfig`

https://docs.cloudposse.com/development/editor-config/

osterman (approved):

PR #76, opened 2018-07-09
Fix template. Link to footnotes broken.

vadim-gleyf:
### What
* Replace \[\]\(\) on \[\]\[\]
* Change releases badge link
* Re-render readme

### Why
* \[\]\(\) style doesn't support variables. [Docs](https://github.com/hugo-2013-07-29/hugo-2013-07-29/blob/master/src/github.com/theplant/blackfriday/upskirtref/Markdown%20Documentation%20-%20Syntax.text#L548)
* Badge has wrong url
* Old template has a bug so avatars links were broken

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Cloud Posse](https://cloudposse.com/logo-300x69.svg)](https://cloudposse.com)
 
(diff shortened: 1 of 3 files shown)


osterman (approved):
Thanks! Good fix. I somehow reintroduced this bug.

PR #79, opened 2018-07-16
Added Local dev tasks

goruha:
## What
* Redo `help` pattern
* Added `aws/shell` target
* Added `WITH_AWS` command allow to run commands under `aws-vault exec`
* Added `chamber/shell` target
* Added `WITH_CHAMBER` command allow to run commands under `chamber exec`
* Added `compose/*` targets
* Added `docker/clean/*` targets
* Added `completion` target

## Why
* Make `help` more useful. To specify help in project that use `build-harness` do
```
export DEFAULT_HELP_TARGET ?= help
## This help screen
help:
	@$(SELF) -s help/all | grep -E "\w(api|compose|worker|redis|deps|help|init|chamber|bash)"
```
* `aws/shell` allow to run commands after `aws-vault` authorization
* `WITH_AWS` simplify doing AWS authorization to  run commands in make files. Ex.: `$(WITH_AWS) aws s3 ls`
* `aws/shell` allow to run commands under `chamber exec`
* `WITH_CHAMBER` simplify access to secrets and run commands in make files. Ex.: `$(WITH_CHAMBER) gomplate ...`
* Target to deal with `docker-compose` for local enviroment
* Targest to cleanup docker trunk
* To make bash completion


aknysh (approved):

modules/aws/Makefile
```diff
@@ -1,7 +1,18 @@
...
+WITH_AWS ?= aws-vault exec $(AWS_PROFILE) --
+
 ## Install aws cli bundle
 aws/install:
 	pip install --no-cache-dir --disable-pip-version-check awscli==$(AWSCLI_VERSION)
 
+## Authorize on aws and run any commands with
+aws/shell: aws/check-shell
+	$(call assert-unset,AWS_VAULT)
+	@$(WITH_AWS) bash --rcfile $(BUILD_HARNESS_PATH)/modules/aws/aws.bash.rc
```
osterman: this is nice

modules/aws/aws.bash.rc
```diff
@@ -0,0 +1,2 @@
+source ~/.bashrc
```
osterman: check if this file first exists before sourcing it

modules/chamber/Makefile
```diff
@@ -2,8 +2,28 @@ CURL ?= curl
...
 CHAMBER ?= chamber
 
+ifdef AWS_VAULT
+	CHAMBER_AUTH = ""
+else
+	CHAMBER_AUTH = $(WITH_AWS)
+endif
+
+WITH_CHAMBER = $(CHAMBER_AUTH) chamber exec $(CHAMBER_SERVICES) --
+CHAMBER_EXPORT = $(CHAMBER_AUTH) chamber export -f dotenv $(CHAMBER_SERVICES) | xargs -I{} echo " {}" | tr '\n' ' '
```
osterman: Can you explain what this command is doing?

modules/chamber/Makefile
```diff
@@ -2,8 +2,28 @@ CURL ?= curl
...
+
+INJECT_CHAMBER ?= $(foreach var, IN_CHAMBER_MODE=true $(shell $(CHAMBER_EXPORT)), $(eval export $(var)))
```
osterman: This part has me concerned. I want to avoid `eval` *at all cost*; developers just don't get it that `Makefile` is a template.

osterman: Also, I don't see where `INJECT_CHAMBER` is used

modules/docker/Makefile
```diff
@@ -1 +1,11 @@
 DOCKER:= $(shell which docker)
+
+DOCKER_CMD ?= docker run -i --rm --network=$(DOCKER_NETWORK)
+SSH_EXEC ?= $(DOCKER) exec -it
```
osterman: I think we should just call this what it is. `DOCKER_EXEC`

modules/docker/Makefile
```diff
@@ -1 +1,11 @@
...
+CONTAINER_SHELL ?= sh
+
+# Connect ssh $host
+define ssh
```
osterman: this should be `docker_exec`; we can describe it in comments like "it's like ssh for docker"

modules/docker/Makefile.cleanup
```diff
@@ -0,0 +1,39 @@
...
+
+
+## Cleanup docker containers.          WARNING!!! IT WILL DELETE ALL UNUSED CONTAINERS
+docker/cleanup/containers:
+	@echo "Cleanup containers"
+	@docker container prune
+	@echo "Done"
+
+## Cleanup docker.                     WARNING!!! IT WILL DELETE ALL UNUSED RESOURCES
+docker/cleanup:
```
osterman: the standard `make` target is `clean`, so i think we should call this `docker/clean`. also, rename the `cleanup` namespace to `clean`

osterman: (standard target names are `all`, `default`, `build`, `clean`, `install`, `deps`)

modules/chamber/Makefile
```diff
@@ -2,8 +2,28 @@ CURL ?= curl
...
 
+## Open secrets and run any commands with
+chamber/shell: chamber/check-shell
+	@if [[ -z "${IN_CHAMBER_MODE}" ]]; then \
+		bash --rcfile $(BUILD_HARNESS_PATH)/modules/chamber/chamber.bash.rc ; \
+	else \
+		$(WITH_CHAMBER) bash --rcfile $(BUILD_HARNESS_PATH)/modules/chamber/chamber.bash.rc ; \
+	fi;
 
+chamber/check-shell: aws/check-shell
```
osterman: Describe what this does

modules/aws/aws.bash.rc
```diff
@@ -0,0 +1,2 @@
...
+PS1=${PS1}"(aws mode)$ "
```
osterman: Consider emitting information about the session so they know the role they assumed:

https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/profile.d/aws-vault.sh#L5

osterman: or simply echo `AWS_PROFILE` here

modules/aws/Makefile
```diff
@@ -1,7 +1,18 @@
 export AWSCLI_VERSION ?= 1.11.185
 
+WITH_AWS ?= aws-vault exec $(AWS_PROFILE) --
+
 ## Install aws cli bundle
 aws/install:
 	pip install --no-cache-dir --disable-pip-version-check awscli==$(AWSCLI_VERSION)
```
osterman, line 7: :(

but i guess it's necessary? thing is `pip` command is not-standard. sometimes it's `pip3`

Makefile.helpers
```diff
@@ -26,18 +26,19 @@ endef
 default:: $(DEFAULT_HELP_TARGET)
 	@exit 0
 
-.PHONY : help
-## This help screen
-help:
-	@$(SELF) help/generate MAKEFILE_LIST="Makefile $(BUILD_HARNESS_PATH)/Makefile.helpers"
+.PHONY : help/short
+## This minimal help screen
+help/short:
```
osterman: I like that you added this target

Makefile.helpers
```diff
@@ -26,18 +26,19 @@ endef
 default:: $(DEFAULT_HELP_TARGET)
 	@exit 0
 
-.PHONY : help
-## This help screen
-help:
```
osterman: Unsure how I feel about this. We show `make help` everywhere in docs, `README.md`, etc. Now it's gone and won't work. I think it should be replaced with:

```
.PHONY : help
## Default help
help:: $(DEFAULT_HELP_TARGET)
  @exit 0
```

modules/completion/Makefile
```diff
@@ -0,0 +1,13 @@
+BASH_COMPLETION_FLAG ?= ~/.make.bash.completion
+
+completion:
```
osterman: I think this should be in `make` module, so `make make/completion`

modules/compose/Makefile
```diff
@@ -0,0 +1,44 @@
...
+DOCKER_CMD ?= docker run -i --rm --network=$(DOCKER_NETWORK)
+
+SSH_EXEC ?= docker exec -it
+CONTAINER_SHELL ?= sh
+
+DOCKER_COMPOSE ?= docker-compose
+
+## Start local dev environment
+compose/up:
+	@$(DOCKER_COMPOSE) up -d
```
osterman: unsure about the `-d` as a default behavior. it's misleading as it masks errors. I've paired with customers and they think it's all fine since `compose/up` exists. Yet a lot of errors are masked. I'm unsure since I don't know which use-case to optimize for.

modules/compose/Makefile
```diff
@@ -0,0 +1,44 @@
+DOCKER_NETWORK ?= $(COMPOSE_PROJECT_NAME)_default
+
+DOCKER_MONITOR_TABLE ?= 'table {{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.NetIO}}\t{{.BlockIO}}'
+
+DOCKER_CMD ?= docker run -i --rm --network=$(DOCKER_NETWORK)
+
+SSH_EXEC ?= docker exec -it
+CONTAINER_SHELL ?= sh
```
osterman: This is declared in a couple of places.

modules/compose/Makefile
```diff
@@ -0,0 +1,44 @@
+DOCKER_NETWORK ?= $(COMPOSE_PROJECT_NAME)_default
+
+DOCKER_MONITOR_TABLE ?= 'table {{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.NetIO}}\t{{.BlockIO}}'
+
+DOCKER_CMD ?= docker run -i --rm --network=$(DOCKER_NETWORK)
+
+SSH_EXEC ?= docker exec -it
```
osterman: This is declared in a couple of places. One will mask the other. Also, see other comments about name of this variable.

modules/docker/Makefile
```diff
@@ -1 +1,11 @@
 DOCKER:= $(shell which docker)
```
osterman: why `:=` ? (add spaces around it)

modules/chamber/Makefile
```diff
@@ -2,8 +2,28 @@ CURL ?= curl
...
+WITH_CHAMBER = $(CHAMBER_AUTH) chamber exec $(CHAMBER_SERVICES) --
+CHAMBER_EXPORT = $(CHAMBER_AUTH) chamber export -f dotenv $(CHAMBER_SERVICES) | xargs -I{} echo " {}" | tr '\n' ' '
+
+INJECT_CHAMBER ?= $(foreach var, IN_CHAMBER_MODE=true $(shell $(CHAMBER_EXPORT)), $(eval export $(var)))
+
 ## Install chamber
 chamber/install: packages/install/chamber
 	@exit 0
 
+## Open secrets and run any commands with
```
osterman: "Start a chamber shell with secrets exported to the environment"

modules/chamber/Makefile
osterman: I don't think this env is needed any more because `INJECT_CHAMBER` is not used anywhere.

modules/completion/Makefile
```diff
@@ -0,0 +1,13 @@
+BASH_COMPLETION_FLAG ?= ~/.make.bash.completion
```
osterman, line 2: This looks more like a `_FILE` than a `_FLAG`

modules/compose/Makefile
```diff
@@ -0,0 +1,44 @@
...
+CONTAINER_SHELL ?= sh
+
+DOCKER_COMPOSE ?= docker-compose
+
+## Start local dev environment
```
osterman: add "(daemonized)"

osterman:
Overall, I really like the direction of this PR, however, lots of feedback =)

osterman (changes requested):
See inline comments

modules/chamber/Makefile
goruha: https://github.com/sportech-dev/local-dev/blob/chamber-support/Makefile#L14

After that make process got access to all secrets from chamber as usual environment variables

modules/chamber/Makefile
goruha: https://github.com/sportech-dev/local-dev/blob/chamber-support/Makefile#L14

modules/aws/Makefile
goruha: As you see I did not add target `aws/install`. 
Probably that should be deprecated because we have `packages/%`?

Makefile.helpers
goruha: If we will add `help` target then 
we can not do this way
https://github.com/sportech-dev/local-dev/blob/chamber-support/Makefile#L21

goruha: Looks like I have idea.

modules/completion/Makefile
goruha: Yea this is the file that perform functionality of flag. 
If it exists - we do not add anything else to `.bashrc` in `completion/install/bash`

modules/chamber/Makefile
goruha: The idea:
Add following line in place you would like to have secrets from chamber:
```
$(INJECT_CHAMBER)
```

That line will export secrets as environment variables 
Where 
* `CHAMBER_EXPORT` export secrets in `.env` format join it into one line separated by space 
* `INJECT_CHAMBER` for each secret (+ IN_CHAMBER_MODE) run `export`

modules/completion/Makefile
goruha: `completion/bash` is better for me
What do you think?

modules/chamber/Makefile
osterman: I just don't get why we we would ever want to do this rather than use `chamber` the way it was destined.

goruha: There are 2 ways to use chamber with make.
1) `make chamber/shell` and then run any make target with secrets.
2) Inject secrets
The minus of the first approach is that outside of `chamber/shell` make targets would not work, because they use secret environment variables. 
Ex.: `make compose/up`

The second solution save us from that problem, it export secrets for any make run.
But it cost additional time to request secrets.

I'd prefer to have ability to use both solutions.

modules/aws/aws.bash.rc
osterman: Indent

modules/chamber/Makefile
osterman: You'll have to show me to sell me.

osterman: I am still not convinced we want/need this. Show me next sync up.

osterman: (I went down this *dangerous* path with Gladly. Our `Makefiles` were only understood by me, and me alone. It does not payoff to be clever with make. My preference would be to document one workflow, not 2)

Makefile.helpers
```diff
@@ -6,7 +6,7 @@ EDITOR ?= vim
 
 SHELL = /bin/bash
 
-DEFAULT_HELP_TARGET ?= help
+HELP_FILTER ?= deps|help|init
```
osterman: `.*`

Makefile.helpers
```diff
@@ -6,7 +6,7 @@ EDITOR ?= vim
 
 SHELL = /bin/bash
 
-DEFAULT_HELP_TARGET ?= help
```
osterman, line 9: If we deprecate this, we need to update dependencies: 

https://github.com/search?q=org%3Acloudposse+DEFAULT_HELP_TARGET&type=Code

README.md
```diff
@@ -117,14 +136,13 @@ Available targets:
   helm/repo/update                    Update repo info
   helm/serve/index                    Build index for serve helm charts
   helm/toolbox/upsert                 Install or upgrade helm tiller 
-  helmfile/install                    Install helmfile
-  help                                This help screen
   help/all                            Display help for all targets
+  help/short                          This minimal help screen
```
osterman: I don't see where `help/short` is defined

modules/aws/Makefile
osterman: Currently, `cloudposse/packages` only handles single-file executables

Makefile.helpers
```diff
@@ -23,21 +23,22 @@ define assert-unset
...
+default:: help
 	@exit 0
 
-.PHONY : help
-## This help screen
+## Help screen
 help:
-	@$(SELF) help/generate MAKEFILE_LIST="Makefile $(BUILD_HARNESS_PATH)/Makefile.helpers"
+	@printf "Available targets:\n\n"
+	@$(SELF) -s help/generate | grep -E "\w($(HELP_FILTER))"
```
osterman, line 33: So I agree having a `helm/generate` based on a filter is nice. However, for simple, non-namespaced makefile targets, this has the opposite effect since there's no way to just say include the local targets from the `./Makefile` without being non-DRY and listing them all in the `HELP_FILTER`

All 100+ terraform modules now output the "simple" `make help` which are the targets from just the `Makefile`. 

This was achieved by `help/generate MAKEFILE_LIST="Makefile $(BUILD_HARNESS_PATH)/Makefile.helpers"`

I think we should have `help/short` run this command. 

Now, I'm not sure which behavior to set to `DEFAULT_HELP_TARGET` so we don't introduce regressions or have to submit hundreds more PRs. This is what that looks like: https://github.com/cloudposse/terraform-aws-ses-lambda-forwarder/blob/create-readme-yaml/docs/targets.md

goruha: Fixed

README.md
```diff
@@ -136,8 +136,9 @@ Available targets:
   helm/repo/update                    Update repo info
   helm/serve/index                    Build index for serve helm charts
   helm/toolbox/upsert                 Install or upgrade helm tiller 
+  help                                Help screen
   help/all                            Display help for all targets
-  help/short                          This minimal help screen
+  help/short                          This help short screen
```
osterman: This short help screen

osterman (changes requested):
Fix wording & rebuild readme

modules/completion/Makefile
```diff
@@ -0,0 +1,10 @@
...
+
+completion/bash:
+	$(SELF) -s $(DEFAULT_HELP_TARGET) | cut -d ' ' -f3  | sed 's/\x1b\[[0-9;]*m//g'
+
+## Install completion script for bash
+completion/install/bash:
+	@[ -f $(BASH_COMPLETION_FLAG) ] || echo "complete -W \"\\\`[ -f Makefile ] && make completion/bash \\\`\" make" >> ~/.bashrc
```
osterman: Does this work in directories that have a `Makefile`, but no `completion/bash` target? I think I ran into this error.

osterman: (basically, does gracefully disable when not in a `build-harness` compatible make environment)

modules/completion/Makefile
osterman: Okay, add a comment above this line to describe what it does.
```
# Flag used to indicate if bash completion rules for make have already been installed
```

modules/completion/Makefile
```diff
@@ -0,0 +1,10 @@
+BASH_COMPLETION_FLAG ?= ~/.make.bash.completion
+
```
osterman, line 3: Annotate this target. 
```
# Target to generate tab completion rules for bash
```

osterman (changes requested):
See updated inline comments.

osterman (approved):

modules/completion/Makefile
goruha: Done

Makefile.helpers
goruha: That feature reverted

Issue #82, opened 2018-07-24
Add `github/push-artifacts` target


osterman:
## what
* Add a new target to upload artifacts from a directory to a github tag/release

Example interface:
`make github/push-artifacts ARTIFACTS=files/ GIT_TAG=0.1.0`

## why
* So we can easily upload artifacts in a codefresh build step

## references

Use one of these projects. Pick the best tool for our use-case.

- https://github.com/aktau/github-release
- https://github.com/tcnksm/ghr

Add the binary package to `cloudposse/packages`

vadim-gleyf:
https://github.com/cloudposse/build-harness/pull/83

PR #83, opened 2018-07-25
Add new target for uploading binaries to release

vadim-gleyf:
### what
* Add new `github/push-artifacts` target

### why
* For uploading artifacts in a codefresh build step


bin/github_push_artifacts.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# PREREQUISITES
+#
+# ghr
+
+TAG=$1
+ARTIFACTS=$2
+GITGUB_TOKEN=$3
+
+ghr -t $GITGUB_TOKEN $TAG $ARTIFACTS
```
osterman: I don't see any benefit with having a shell script here. Use shell scripts when there is complicated bash logic/flow control.  Use inline commands in make targets when there's just a one-liner.

osterman:
This PR should be depended on another one in packages, but I don't see any here: 

https://github.com/cloudposse/packages/pulls

osterman (changes requested):
See inline comments

modules/github/Makefile.release
```diff
@@ -19,3 +19,9 @@ github/download-public-release:
 	$(call assert-set,VERSION)
 	$(call assert-set,OUTPUT)
 	$(BUILD_HARNESS_PATH)/bin/github_download_public_release.sh $(VERSION) $(OUTPUT)
+
```
osterman, line 22: Add a target description. e.g.
```
## Push all release artifacts to GitHub
```

modules/github/Makefile.release
```diff
@@ -19,3 +19,9 @@ github/download-public-release:
...
+github/push-artifacts:
```
osterman: Add a package dependency here:

```
github/push-artifacts: packages/install/ghr
```

modules/github/Makefile.release
```diff
@@ -19,3 +19,10 @@ github/download-public-release:
 	$(call assert-set,VERSION)
 	$(call assert-set,OUTPUT)
 	$(BUILD_HARNESS_PATH)/bin/github_download_public_release.sh $(VERSION) $(OUTPUT)
+
+## Push all release artifacts to GitHub
+github/push-artifacts: packages/install/ghr
+	$(call assert-set,GIT_TAG)
+	$(call assert-set,ARTIFACTS)
+	$(call assert-set,GITHUB_TOKEN)
+	@ghr -t $(GITHUB_TOKEN) $(GIT_TAG) $(ARTIFACTS)
```
osterman: Can `ghr` work without passing the token as a parameter? E.g. can it look in the environment? Would rather not pass it on the command line if we can avoid it.

osterman (commented):
See inline comments

goruha (changes requested):
Update package version to use https://github.com/cloudposse/packages/pull/27

osterman (changes requested):
See comment by @goruha

vadim-gleyf:
#### Depend on 
- https://github.com/cloudposse/packages/pull/27

modules/github/Makefile.release
```diff
@@ -19,3 +19,11 @@ github/download-public-release:
 	$(call assert-set,VERSION)
 	$(call assert-set,OUTPUT)
 	$(BUILD_HARNESS_PATH)/bin/github_download_public_release.sh $(VERSION) $(OUTPUT)
+
+## Push all release artifacts to GitHub
+## To use ghr, you need a GitHub API token.
```
osterman: Target annotations `##` must be on one-line and cannot repeat. Also, they must immediately precede the target name. This is due to how it's implemented with `awk`: https://github.com/cloudposse/build-harness/blob/master/Makefile.helpers#L39-L51

Instead, notes to developers should just use a single `#`. See other examples in makefiles.

modules/github/Makefile.release
```diff
@@ -19,3 +19,11 @@ github/download-public-release:
 	$(call assert-set,VERSION)
 	$(call assert-set,OUTPUT)
 	$(BUILD_HARNESS_PATH)/bin/github_download_public_release.sh $(VERSION) $(OUTPUT)
+
+## Push all release artifacts to GitHub
```
osterman: Perhaps write: 
```
## Push all release artifacts to GitHub (Required: `GITHUB_TOKEN`)
```

osterman (changes requested):
See inline comments

goruha (approved):

osterman (approved):

osterman:
https://github.com/cloudposse/build-harness/issues/82

PR #84, opened 2018-07-26
use predefined value of $(MAKE) for SELF

tamsky:
this allows one to run `make -n` and have it carry through correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 export BUILD_HARNESS_PATH ?= $(shell 'pwd')
 export OS ?= $(shell uname -s | tr '[:upper:]' '[:lower:]')
-export SELF ?= make
+export SELF ?= $(MAKE)
 export PATH := $(BUILD_HARNESS_PATH)/vendor:$(PATH)
 export DOCKER_BUILD_FLAGS ?=
 


osterman (approved):

goruha (changes requested):
We user `$(SELF)`
Recommend to close this PR

osterman:
@goruha recommend closing or recommend changing to use `SELF`? =)

https://github.com/cloudposse/build-harness/blob/master/Makefile#L3

Perhaps, `SELF` should be updated to default to `$(MAKE)` instead of `make`

tamsky:
> Perhaps, SELF should be updated to default to $(MAKE) instead of make

this.
but it certainly makes me wonder why have SELF in the first place if $(MAKE) is right there.

I can update existing recipe usage of `@make` -> `$(SELF)` and update `SELF = $(MAKE)`, but maybe it's worth examining why SELF is there.

osterman:
I think `SELF` could be an artifact from long ago and could possibly be replaced with `MAKE`. I am just concerned about the implications since this is used by a lot of projects. Suggest a two phased approach. Change `SELF = $(MAKE)` and then we can replace it later. 

At some point, we needed to have `make -f Makefile` because had problems using `MAKE` recursively.  I cannot recall why or even if that was in this `build-harness` or [another one](https://github.com/sagansystems/build-harness/blob/master/Makefile#L4) I used years ago.

tamsky:
> Suggest a two phased approach. Change SELF = $(MAKE) and then we can replace it later.

updated this PR to implement phase one.

goruha (approved):

osterman:
Travis is failing since it won't export ENVs for untrusted builds. Going to merge regardless.

Issue #89, opened 2018-08-02
Feature Flag: Silence Warnings


osterman:
## what
* Add a feature flag to make it possible to disable warnings

```
//build-harness/modules/git/bootstrap.Makefile:2: disabling git bootstrapping
//build-harness/modules/helm/Makefile.repo:15: unable to extrapolate current BRANCH
//build-harness/modules/semver/Makefile:58: disabled semver processing
//build-harness/modules/travis/Makefile:12: unable to extrapolate current BRANCH
//build-harness/modules/git/bootstrap.Makefile:2: disabling git bootstrapping
//build-harness/modules/helm/Makefile.repo:15: unable to extrapolate current BRANCH
//build-harness/modules/semver/Makefile:58: disabled semver processing
//build-harness/modules/travis/Makefile:12: unable to extrapolate current BRANCH
```

## why
* Warnings are great for debugging, but scary for users

PR #90, opened 2018-08-08
Silence warnings without DEBUG flag

vadim-gleyf:
## what
* Add a feature flag to make it possible to disable warnings

## why
* Warnings are great for debugging, but scary for users

## references
- https://github.com/cloudposse/build-harness/issues/89

### DIFF
--- a/modules/git/bootstrap.Makefile
+++ b/modules/git/bootstrap.Makefile
@@ -1,6 +1,6 @@
-export DEBUG ?= false
+export DEBUG ?= disabled
 
-ifeq ($(DEBUG),true)
+ifeq ($(DEBUG),enabled)
 ifeq ($(wildcard .git),)
   $(warning disabling git bootstrapping)
 else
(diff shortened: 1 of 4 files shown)


osterman:
LGTM! Please output with debug enabled and debug disabled.

osterman (changes requested):
See comments. Post output in comments from test runs.

modules/git/bootstrap.Makefile
```diff
@@ -27,3 +28,4 @@ else
 endif
 
 endif
+endif
```
aknysh: is it correct that if `d` flag is not set, we skip the entire file?
do we want to disable just the warning?

```
ifeq ($(wildcard .git),)
  ifneq (,$(findstring d,$(MAKEFLAGS)))
      $(warning disabling git bootstrapping)
  endif
else
.....
endif
```

aknysh (commented):
see comments

modules/git/bootstrap.Makefile
osterman: @aknysh good catch! That does look like the case. @vadim-hleif We only want to disable the warnings, not the logic =)

aknysh (approved):

vadim-gleyf:
capture

- Just copied part of code for demonstrating

osterman (approved):

Issue #96, opened 2018-08-23
[packages/install] only install package if not already installed


osterman:
## what
* We should check to see if the package is already installed (e.g. in path)
* Add `packages/reinstall/%` target to reinstall a package

## why
* Reinstalling is slow

PR #98, opened 2018-08-24
Change packages/install target behavior

vadim-gleyf:
## what
* Now package wouldn't be installed if it exist in install path
* Add new target packages/reinstall (old packages/install target behavior)
* Regenerate README

## why 
* It's tedious to reinstall dependencies every time.

## references
* https://github.com/cloudposse/build-harness/issues/96

### DIFF
--- a/modules/packages/Makefile
+++ b/modules/packages/Makefile
@@ -17,11 +17,20 @@ packages/install:
 	  rm -rf $(PACKAGES_PATH); \
 	  git clone -c advice.detachedHead=false --depth=1 -b $(PACKAGES_VERSION) https://github.com/cloudposse/packages.git $(PACKAGES_PATH); \
 	  rm -rf $(PACKAGES_PATH)/.git; \
-	fi
+	fi 
 
 ## Install package (e.g. helm, helmfile, kubectl)
 packages/install/%: packages/install
-	@make -C $(PACKAGES_PATH)/install $(subst packages/install/,,$@)
+	$(eval $@_package=$(subst packages/install/,,$@)) 
+	@if [ ! $(shell type $(INSTALL_PATH)/$($@_package) &>/dev/null && echo $?) ]; then \
+		make -C $(PACKAGES_PATH)/install $($@_package); \
+	else \
+		echo "Package $($@_package) already installed"; \
+	fi;
+
+## Reinstall package (e.g. helm, helmfile, kubectl)
+packages/reinstall/%: packages/install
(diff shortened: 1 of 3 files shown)


modules/packages/Makefile
```diff
@@ -17,11 +17,20 @@ packages/install:
...
 	  git clone -c advice.detachedHead=false --depth=1 -b $(PACKAGES_VERSION) https://github.com/cloudposse/packages.git $(PACKAGES_PATH); \
 	  rm -rf $(PACKAGES_PATH)/.git; \
-	fi
+	fi 
 
 ## Install package (e.g. helm, helmfile, kubectl)
 packages/install/%: packages/install
-	@make -C $(PACKAGES_PATH)/install $(subst packages/install/,,$@)
+	$(eval $@_package=$(subst packages/install/,,$@)) 
+	@if [ ! $(shell type $(INSTALL_PATH)/$($@_package) &>/dev/null && echo $?) ]; then \
```
osterman: Let's avoid using `$(shell....)`; use `$$(...)` instead.

modules/packages/Makefile
```diff
@@ -17,11 +17,20 @@ packages/install:
 	  rm -rf $(PACKAGES_PATH); \
 	  git clone -c advice.detachedHead=false --depth=1 -b $(PACKAGES_VERSION) https://github.com/cloudposse/packages.git $(PACKAGES_PATH); \
 	  rm -rf $(PACKAGES_PATH)/.git; \
-	fi
+	fi 
 
 ## Install package (e.g. helm, helmfile, kubectl)
 packages/install/%: packages/install
-	@make -C $(PACKAGES_PATH)/install $(subst packages/install/,,$@)
+	$(eval $@_package=$(subst packages/install/,,$@))
```
osterman: Let's not use `eval`; shouldn't be necessary

osterman: let's do something instead like this:

```
@if ! which $(subst packages/install/,,$@) 2>/dev/null; then \
  make -C $(PACKAGES_PATH)/install; \
 fi
```

osterman: Basically, if app is not found in `PATH` (which we confirm with `which`), then install

osterman: The other option is to do:
```
@if [ ! -x $(INSTALL_PATH)/$(subst packages/install/,,$@) ]; then \
....
fi
```

osterman (changes requested):
See comments

osterman (approved):

PR #100, opened 2018-08-30
Dynamic datasources support

### LINKED ISSUE #87: Update Readme Generator to use New Gomplate Inline Datasources
## what
* New version gomplate of gomplate supports datasources to be defined inline

## why
* We should use this to simplify invocation

## references
- <https://github.com/hairyhenderson/gomplate/issues/349>
- https://github.com/hairyhenderson/gomplate/pull/373

vadim-gleyf:
## what
* Now use url from arg `datasource` if it passed
* Regenerate readme

## why
* https://github.com/cloudposse/build-harness/issues/87

### Depend on [#55](https://github.com/cloudposse/packages/pull/55)
* Need to bump `packages` [version](https://github.com/cloudposse/build-harness/blob/master/modules/packages/Makefile#L2) when it merged 

resolve: #87


osterman (changes requested):
The objective is to move all data source definitions into the readme template. That way, we can simply pass —file and —out and it works. No obligation on the caller to provide data sources.

templates/README.md
```diff
@@ -1,5 +1,8 @@
 <!-- This file was automatically generated by the `build-harness`. Make all changes to `README.yaml` and run `make readme` to rebuild this file. -->
-
+<!-- 
+{{ defineDatasource "config" .Env.DATASOURCE }}
```
osterman: Use {{- ... -}} instead of the HTML comments

modules/readme/Makefile
```diff
@@ -3,6 +3,9 @@ export README_FILE ?= README.md
 export README_YAML ?= README.yaml
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md
 export README_TEMPLATE_YAML ?= $(BUILD_HARNESS_PATH)/templates/README.yaml
+export USED_DATASOURCE ?= $(if $(datasource),$(datasource),$(shell pwd)/$(README_YAML))
```
osterman: What is this for?

osterman (changes requested):

modules/readme/Makefile
vadim-gleyf: @osterman  For correct output 
`@echo "Generated $(README_FILE) from $(README_TEMPLATE_FILE) using data from $(USED_DATASOURCE)"`

templates/README.md
```diff
@@ -1,5 +1,6 @@
 <!-- This file was automatically generated by the `build-harness`. Make all changes to `README.yaml` and run `make readme` to rebuild this file. -->
-
+{{ defineDatasource "config" .Env.DATASOURCE }}
```
osterman: This is what I meant we could do https://github.com/cloudposse/geodesic/blob/master/rootfs/templates/kops/default.yaml#L57

modules/readme/Makefile
```diff
@@ -3,6 +3,9 @@ export README_FILE ?= README.md
...
+export DATASOURCE ?= $(if $(datasource),$(datasource),$(README_YAML))
```
osterman: where is lowercase `datasource` ever defined?

modules/readme/Makefile
```diff
@@ -29,7 +32,5 @@ readme/lint:
 ## Create README.md by building it from README.yaml
 readme/build: $(README_DEPS)
 	@gomplate --file $(README_TEMPLATE_FILE) \
-		--out $(README_FILE) \
-		--datasource config=$(README_YAML) \
-		--datasource includes='file://$(shell pwd)/?type=text/plain'
-	@echo "Generated $(README_FILE) from $(README_TEMPLATE_FILE) using data from $(README_TEMPLATE_YAML)"
+		--out $(README_FILE) 
+	@echo "Generated $(README_FILE) from $(README_TEMPLATE_FILE) using data from $(USED_DATASOURCE)"
```
osterman: What's wrong with using `README_TEMPLATE_YAML`? I don't understand why we need a new variable.

templates/README.md
```diff
@@ -1,5 +1,6 @@
 <!-- This file was automatically generated by the `build-harness`. Make all changes to `README.yaml` and run `make readme` to rebuild this file. -->
-
+{{ defineDatasource "config" .Env.DATASOURCE | regexp.Replace ".*" ""}}
```
osterman: add a space between the quote and closing brace. `"}}`

templates/README.md
```diff
@@ -1,5 +1,6 @@
...
+{{ defineDatasource "includes" .Env.INCLUDES | regexp.Replace ".*" ""}}
```
osterman: add a space between the quote and closing brace. `"}}`

modules/readme/Makefile
vadim-gleyf: @osterman for overriding default datasource (`README.yaml` in root folder)
image

modules/readme/Makefile
vadim-gleyf: @osterman due possible overriding, it is not necessary?

modules/readme/Makefile
osterman: Is there a way we could just get rid of these envs and define it in the template directly?

osterman (changes requested):
Can we get rid of the envs for the template and just hard code something in the template? My hope is to make the template almost entirely self-contained.

modules/readme/Makefile
vadim-gleyf: @osterman now if no any env passed, behaviour will be as earlier, will be used `README.yaml` from root project folder. 
This env just for possibility to change datasource, it isn't use case?

vadim-gleyf: @osterman move datasource definition from makefile, but gomplate doesn't support inline shell execution, 
`export INCLUDES ?= $(file://$(shell pwd)/?type=text/plain)`
still here

modules/readme/Makefile
```diff
@@ -3,6 +3,8 @@ export README_FILE ?= README.md
 export README_YAML ?= README.yaml
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md
 export README_TEMPLATE_YAML ?= $(BUILD_HARNESS_PATH)/templates/README.yaml
+export USED_DATASOURCE ?= $(if $(datasource),$(datasource),$(shell pwd)/$(README_YAML))
```
osterman: Where does is `$(datasource)` defined? I would expect to see `datasource =` somewhere.

osterman: The user can just specify `make readme README_TEMPLATE_YAML=foobar.yaml` if they want to override, so unsure what `datasource` gives us.

osterman (changes requested):
See comments

modules/readme/Makefile
```diff
@@ -2,8 +2,7 @@ export README_LINT ?= $(TMP)/README.md
 export README_FILE ?= README.md
 export README_YAML ?= README.yaml
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md
-export README_TEMPLATE_YAML ?= $(BUILD_HARNESS_PATH)/templates/README.yaml
-export USED_DATASOURCE ?= $(if $(datasource),$(datasource),$(shell pwd)/$(README_YAML))
+export README_TEMPLATE_YAML := $(if $(findstring http,$(README_YAML)),$(README_YAML),$(BUILD_HARNESS_PATH)/templates/$(README_YAML))
 export INCLUDES ?= $(file://$(shell pwd)/?type=text/plain)
```
osterman: Rename to `README_INCLUDES` so we don't overload this env.

osterman (changes requested):
Rename `INCLUDES`

modules/readme/Makefile
```diff
@@ -3,7 +3,7 @@ export README_FILE ?= README.md
 export README_YAML ?= README.yaml
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md
 export README_TEMPLATE_YAML := $(if $(findstring http,$(README_YAML)),$(README_YAML),$(BUILD_HARNESS_PATH)/templates/$(README_YAML))
-export INCLUDES ?= $(file://$(shell pwd)/?type=text/plain)
+export FILES ?= $(file://$(shell pwd)/?type=text/plain)
```
osterman: No, rename to `README_INCLUDES` - the objective is to "namespace" environment variables related to the readme generation so they don't conflict. Generic names like `INCLUDES` or `FILES` are likely to "overload" other envs.

osterman (changes requested):
See comments

osterman (dismissed):

modules/readme/Makefile
```diff
@@ -2,7 +2,8 @@ export README_LINT ?= $(TMP)/README.md
 export README_FILE ?= README.md
 export README_YAML ?= README.yaml
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md
-export README_TEMPLATE_YAML ?= $(BUILD_HARNESS_PATH)/templates/README.yaml
+export README_TEMPLATE_YAML := $(if $(findstring http,$(README_YAML)),$(README_YAML),$(BUILD_HARNESS_PATH)/templates/$(README_YAML))
```
osterman, line 5: Can you elaborate on why we have findstring here?

osterman (commented):
Explain need for `findstring`

modules/readme/Makefile
vadim-gleyf: @osterman to check is datasource external or not

osterman (approved):

PR #104, opened 2018-09-07
[gitleaks] add support

### LINKED ISSUE #91: Add Code Credential Security Scanner
## what
* Add target that checks for scanning code for credential leaks

## why
* Ensure we don't accidentally leak secrets (e.g. git credentials, aws keys) in any of our projects

## references
- <https://github.com/cloudposse/packages/issues/40>
- <https://github.com/cloudposse/copyright-header/issues/49>

alebabai:
## what
add `gitleaks` module to scan code for credential leaks

## why
closes #91

### DIFF
--- a/modules/gitleaks/Makefile
+++ b/modules/gitleaks/Makefile
@@ -4,6 +4,7 @@ GITLEAKS ?= gitleaks
 
 ## Install gitleaks
 gitleaks/install: packages/install/gitleaks
+	@exit 0
 
 ## Scan current repository
 gitleaks/scan:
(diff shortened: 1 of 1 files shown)


modules/gitleaks/Makefile
```diff
@@ -0,0 +1,10 @@
+export GITLEAKS_VERSION ?= 1.2.0
+
+GITLEAKS ?= gitleaks
+
+## Install gitleaks
+gitleaks/install: packages/install/gitleaks
+
+## Scan current repository
+gitleaks/scan:
```
osterman: should this call `gitleaks/scan: gitleaks/install`?

alebabai: i don't think so, because modules like `helm` or `chamber` also have install target, but didn't invoke it in other targets

osterman (approved):

PR #109, opened 2018-09-19
[modules/go] re-write go vet target

### LINKED ISSUE #29: Fix CI/CD
## what
* Builds are failing
* See https://travis-ci.org/cloudposse/github-authorized-keys/builds/402672478#L819

## why
* I don't know

## references
* https://github.com/cloudposse/github-authorized-keys/pull/28

alebabai:
## what
use `go too vet $dir` to vet code

## why
`go vet` command require:
> named files must all be in one directory; have ...

Should fix https://github.com/cloudposse/github-authorized-keys/issues/29

### DIFF
--- a/modules/go/Makefile.style
+++ b/modules/go/Makefile.style
@@ -6,7 +6,7 @@ go/lint: $(GO) go/vet
 ## Vet code
 go/vet: $(GO)
 	$(call assert-set,GO)
-	$(GO) tool vet -v $(CURDIR)
+	find . ! -path "*/vendor/*" ! -path "*/.glide/*" -type f -name '*.go' | xargs $(GO) tool vet -v
 
 ## Format code according to Golang convention
 go/fmt: $(GO)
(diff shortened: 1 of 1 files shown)


modules/go/Makefile.style
```diff
@@ -6,7 +6,7 @@ go/lint: $(GO) go/vet
 ## Vet code
 go/vet: $(GO)
 	$(call assert-set,GO)
-	find . ! -path "*/vendor/*" ! -path "*/.glide/*" -type f -name '*.go' | xargs $(GO) vet -v
+	$(GO) tool vet -v $(CURDIR)
```
osterman: Will this go recursively? E.g. projects have go code in multiple folders.

osterman (commented):
See comment

modules/go/Makefile.style
alebabai: yeap, it recursively descends the directory, vetting each package it finds.
https://golang.org/cmd/vet/#hdr-Using_vet_directly

osterman:
But we need to exclude the vendor folder since we have to control over the formatting of 3rd party code. Will this automatically exclude that?

alebabai:
@osterman hm, you're right.
i'll bring old solution with find, but using go tool vet

osterman (approved):

PR #112, opened 2018-10-01
[helm.toolbox.upsert]: tiller rbac support

alebabai:
## what
* create service-account and clusterrolebinding for helm tiller (if not exists)
* pass service-account to helm init

## why
to support RBAC on helm level

### DIFF
--- a/bin/helm_toolbox.sh
+++ b/bin/helm_toolbox.sh
@@ -45,16 +45,20 @@ function upsert() {
     if [ "$helm_version" != "$tiller_version" ]; then
         info "Helm version: $helm_version, differs with tiller version: ${tiller_version:-'not installed'}"
         info "Upgrarding tiller to $helm_version"        
-        local tiller_serviceaccount=tiller
-        local tiller_serviceaccount_exists=$(kubectl get serviceaccount -n kube-system --ignore-not-found=true --request-timeout=${TIMEOUT}s -o name $tiller_serviceaccount 2> /dev/null)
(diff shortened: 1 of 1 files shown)


bin/helm_toolbox.sh
```diff
@@ -44,8 +44,17 @@ function upsert() {
     local tiller_version=$(helm version --server --short --tiller-connection-timeout $TIMEOUT 2> /dev/null | grep -Eo "v[0-9]+\.[0-9]+\..+")
     if [ "$helm_version" != "$tiller_version" ]; then
         info "Helm version: $helm_version, differs with tiller version: ${tiller_version:-'not installed'}"
-        info "Upgrarding tiller to $helm_version"
-        helm init --upgrade --force-upgrade --wait > $STDOUT
+        info "Upgrarding tiller to $helm_version"        
+        local tiller_serviceaccount=tiller
+        local tiller_serviceaccount_exists=$(kubectl get serviceaccount -n kube-system --ignore-not-found=true --request-timeout=${TIMEOUT}s -o name $tiller_serviceaccount 2> /dev/null)
+        local tiller_clusterrolebinding_exists=$(kubectl get clusterrolebinding --ignore-not-found=true --request-timeout=${TIMEOUT}s -o name tiller-cluster-rule 2> /dev/null)
```
osterman: Shouldn't we only do this if RBAC enabled?

osterman (changes requested):
See comment

bin/helm_toolbox.sh
alebabai: Yeap, you're right.
I've used RBAC_ENABLED env for it.

alebabai:
verified on testing cluster

osterman (approved):

PR #139, opened 2019-01-29
Export Environment Variables for Codefresh Docker Tags

maximmi:
## What
* exporting some new variables to use as `Docker` tags for `Codefresh` push step. going to use instead `travis_docker_tag_and_push.sh`

## Why
* Because of migrating to `Codefresh`. Now we can use all new variables as `Docker` tags in a single push step, without conditions (by branch, PR, release etc). So only proper variables will be defined.

### DIFF
--- a/modules/codefresh/Makefile
+++ b/modules/codefresh/Makefile
@@ -12,7 +12,12 @@ codefresh/trigger/webhook:
 	 --data-binary \
 		'{"serviceId":"$(CF_TRIGGER_SERVICE_ID)","type":"$(CF_TRIGGER_TYPE)","repoOwner":"$(CF_TRIGGER_REPO_OWNER)","branch":"$(CF_TRIGGER_BRANCH)","repoName":"$(CF_TRIGGER_REPO_NAME)"}'
 
-## DEPRECATED!!! Export codefresh additional envvars
+## Export codefresh additional envvars
 codefresh/export:
 	$(call assert-set,CF_BUILD_TIMESTAMP)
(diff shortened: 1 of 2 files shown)


aknysh (changes requested):
~rename `bootstrap.Makefile` file to `Makefile.bootstrap` - this is a convention to name Makefiles~

modules/codefresh/Makefile
```diff
@@ -12,7 +12,12 @@ codefresh/trigger/webhook:
 	 --data-binary \
 		'{"serviceId":"$(CF_TRIGGER_SERVICE_ID)","type":"$(CF_TRIGGER_TYPE)","repoOwner":"$(CF_TRIGGER_REPO_OWNER)","branch":"$(CF_TRIGGER_BRANCH)","repoName":"$(CF_TRIGGER_REPO_NAME)"}'
 
-## DEPRECATED!!! Export codefresh additional envvars
+## Export codefresh additional envvars
 codefresh/export:
 	$(call assert-set,CF_BUILD_TIMESTAMP)
 	@echo "CF_BUILD_UNIX_TIMESTAMP=$(shell expr $(CF_BUILD_TIMESTAMP) / 1000)"
+	@echo "IS_CODEFRESH=$(IS_CODEFRESH)"
+	@echo "CICD_BUILD_ID=$(CICD_BUILD_ID)"
```
osterman, line 20:
```suggestion
	@echo "DOCKER_TAG_BUILD_ID=$(DOCKER_TAG_BUILD_ID)"
```

osterman: I'm also ok with `CF_TAG_BUILD_ID`

PR #140, opened 2019-02-04
Allow an optional IMAGE_VERSION to be added for Helm builds

lhw:
In some cases the TARGET_VERSION can and should differ from the IMAGE_VERSION.
This merge request would allow for an optional IMAGE_VERSION parameter to be used for the helm build.

### DIFF
--- a/modules/helm/Makefile.build
+++ b/modules/helm/Makefile.build
@@ -77,7 +77,7 @@ helm/chart/build/package:
 	$(call assert-set,TARGET_VERSION)
 	$(call assert-set,HELM_PACKAGE_PATH)
 	$(eval IMAGE_VERSION ?= $(TARGET_VERSION))
-	@echo "For back compatibility pinning image.tag to $(IMAGE_VERSION)"
+	@echo "For backward compatibility pinning image.tag to $(IMAGE_VERSION)"
 	@set -o pipefail; yq write --inplace $(HELM_PACKAGE_PATH)$(CHART_NAME)/values.yaml image.tag $(IMAGE_VERSION) | head -1
 	@echo "Pack $(CHART_NAME) with version $(TARGET_VERSION)"
 	@$(HELM) package  \
--- a/modules/helm/Makefile.build
+++ b/modules/helm/Makefile.build
@@ -71,13 +71,14 @@ helm/chart/build/deps:
 	@echo "Fetch dependencies for $(HELM_PACKAGE_PATH)$(CHART_NAME)"
 	@$(HELM) dependency build --debug $(HELM_PACKAGE_PATH)$(CHART_NAME)
 
-# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH
+# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH optionally pinning it with IMAGE_VERSION
 helm/chart/build/package:
 	$(call assert-set,CHART_NAME)
 	$(call assert-set,TARGET_VERSION)
 	$(call assert-set,HELM_PACKAGE_PATH)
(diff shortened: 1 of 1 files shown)


modules/helm/Makefile.build
```diff
@@ -71,13 +71,14 @@ helm/chart/build/deps:
...
-# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH
+# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH optionally pinning it with IMAGE_VERSION
 helm/chart/build/package:
 	$(call assert-set,CHART_NAME)
 	$(call assert-set,TARGET_VERSION)
 	$(call assert-set,HELM_PACKAGE_PATH)
-	@echo "For back compatibility pinning image.tag to $(TARGET_VERSION)"
-	@set -o pipefail; yq write --inplace $(HELM_PACKAGE_PATH)$(CHART_NAME)/values.yaml image.tag $(TARGET_VERSION) | head -1
+	$(eval IMAGE_VERSION ?= $(TARGET_VERSION))
+	@echo "For back compatibility pinning image.tag to $(IMAGE_VERSION)"
```
aknysh:
```suggestion
	@echo "For backward compatibility pinning image.tag to $(IMAGE_VERSION)"
```

modules/helm/Makefile.build
```diff
@@ -71,13 +71,14 @@ helm/chart/build/deps:
...
 
-# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH
+# Create a helm package called CHART_NAME and store it in HELM_PACKAGE_PATH optionally pinning it with IMAGE_VERSION
 helm/chart/build/package:
 	$(call assert-set,CHART_NAME)
 	$(call assert-set,TARGET_VERSION)
 	$(call assert-set,HELM_PACKAGE_PATH)
-	@echo "For back compatibility pinning image.tag to $(TARGET_VERSION)"
-	@set -o pipefail; yq write --inplace $(HELM_PACKAGE_PATH)$(CHART_NAME)/values.yaml image.tag $(TARGET_VERSION) | head -1
+	$(eval IMAGE_VERSION ?= $(TARGET_VERSION))
```
goruha, line 79: @lhw is it possible to avoid `eval` ?

goruha (changes requested):

modules/helm/Makefile.build
lhw: My aim here was to do as few changes as possible to make this change happen. As far as I can tell there is no shorter way to do this in Makefiles. You could of course duplicate the lines and do a conditional run. But from my experience this is far from the worst use-case for eval()

osterman: What if we deferred to bash instead?
```
set -o pipefail; yq write --inplace $(HELM_PACKAGE_PATH)$(CHART_NAME)/values.yaml image.tag $${IMAGE_VERSION:-TARGET_VERSION} | head -1
```

osterman: the `:-` operator in the bash variable interpolation will then set the default value if one is not set. 

We like to avoid `eval` at all cost in `Makefiles` because the flow of is not linear.

lhw: I mean sure. But that does create an external dependency where none is needed. I am not all that invested in this change. I just did a quick fix for a co-worker who wanted the option. If you think that eval() is an issue feel free to change it.

osterman: Thanks @lhw. Not trying to be difficult =) Just rather than evaluate on a case-by-case basis whether or not `eval` makes sense in some particular use-case, we outright avoid it altogether. It reduces decisions.

https://docs.cloudposse.com/tools/make/best-practices/

PR #144, opened 2019-04-09
[codefresh-sync] Leave TMP_DIR around for future use

Nuru:
## what
Make Codefresh re-create $TMP_DIR after deleting it

## why
Future builds may need it.


osterman:
Since I think the root of our other problem was addressed, going to close this since I don't know the impact

PR #147, opened 2019-04-14
Add a note about pinning module dependencies

aknysh:
## what
* Add a note about pinning module dependencies

## why
* Users should always do module version pinning for reproducibility and visibility
* Users should pin to a release and not use the `master` branch because there could be breaking changes between releases
* Some users either forget to pin at all, or pin to `master` in production code
* Just a friendly reminder
* Will be included in every terraform module README

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -67,11 +67,13 @@ We literally have [*hundreds of terraform modules*][terraform_modules] that are 
 {{ if has (ds "config") "usage" }}
 ## Usage
 
+{{ if (file.Exists "main.tf") }}
 **IMPORTANT:** The example shows pinning to the `master` branch. 
 This is a reminder that you need to pin your dependencies for reproducibility and visibility.
 Explicit versioning allows knowing exactly which version of each dependency is used at any time.
 However, in your code do not pin to `master` because there may be breaking changes between releases.
 Instead pin to the release tag (e.g. `?ref=tags/x.y.z`) of one of our [latest releases](https://github.com/cloudposse/{{(ds "config").name}}/releases).
+{{end}}
 
 {{ (ds "config").usage -}}
 {{ end }}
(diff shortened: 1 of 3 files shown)


templates/README.md
```diff
@@ -67,6 +67,12 @@ We literally have [*hundreds of terraform modules*][terraform_modules] that are
 {{ if has (ds "config") "usage" }}
 ## Usage
```
osterman, line 69: We should only do this for terraform projects. See example in this file where we check for `main.tf`

templates/README.md
```diff
@@ -67,6 +67,12 @@ We literally have [*hundreds of terraform modules*][terraform_modules] that are
...
 
+**IMPORTANT:** The example shows pinning to the `master` branch.
```
osterman: This overall feels much wordier. Did you not like what we had before?

> **IMPORTANT:** The `master` branch is used in `source` just as an example. In your code, do not pin to `master` because there may be breaking changes between releases. Instead pin to the release tag (e.g. `?ref=tags/x.y.z`) of one of our [latest releases](https://github.com/cloudposse/terraform-aws-rds-cluster/releases).

aknysh: ok, reverted back.

I added more text because I felt it should describe three different things:
1. You should always do version pinning 
2. You should not pin to master
3. But we show pinning to master because we don't want to update all examples every time we update modules

But I agree less words is better.

templates/README.md
```diff
@@ -67,6 +67,11 @@ We literally have [*hundreds of terraform modules*][terraform_modules] that are
 {{ if has (ds "config") "usage" }}
 ## Usage
 
+{{ if (file.Exists "main.tf") }}
+**IMPORTANT:** The `master` branch is used in `source` just as an example. In your code, do not pin to `master` because there may be breaking changes between releases.
+Instead pin to the release tag (e.g. `?ref=tags/x.y.z`) of one of our [latest releases](https://github.com/cloudposse/{{(ds "config").name}}/releases).
```
osterman: I think we have a setting for repo name

osterman (changes requested):

osterman (approved):

PR #152, opened 2019-06-04
0.12 support for terraform-docs

osterman:
## what
* Add a hack to support terraform `0.12` with `terraform-docs`
* Use `awk` to generate variables and outputs suitable for `terraform-docs`

## why
* `terraform-docs` does not yet support HCL2

## references
* https://github.com/segmentio/terraform-docs/issues/62


aknysh (approved):

antonbabenko:
Thanks a lot for this solution!

It works for all cases I have, but not when `default` value looks like this:

```hcl
variable "database_outbound_acl_rules" {
  description = "Database subnets outbound network ACL rules"
  type        = list(map(string))

  default = [
    {
      rule_number = 100
      rule_action = "allow"
      from_port   = 0
      to_port     = 0
      protocol    = "-1"
      cidr_block  = "0.0.0.0/0"
    },
  ]
}
```

Do you know how to improve awk script to handle this?

PR #156, opened 2019-06-18
Fix processing of occurances of 'description' and null values

cytopia:
# Fix processing of occurances of 'description' and null values

This PR addresses the following:

* gnu/bsd awk compatibility
* handle occurances of the word 'description' properly
* transform `default = null` to `default = "null"`

### DIFF
--- a/bin/terraform-docs.awk
+++ b/bin/terraform-docs.awk
@@ -4,16 +4,18 @@
 # https://github.com/segmentio/terraform-docs/issues/62
 
 {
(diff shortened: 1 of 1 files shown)


antonbabenko (approved):
Looks good, and it already works as part of [v1.15.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.15.0) of pre-commit-terraform hooks. :)

aknysh (approved):

PR #157, opened 2019-06-30
Fix: Do not double quote TF >= 0.12 legacy quoted types

cytopia:
## Do not double quote TF >= 0.12 legacy quoted types

This PR addresses another issue with double-double quoting legacy types. This happens if you still double quote `string`, `list` and `map` in Terraform >= 0.12.

```
Error: Invalid legacy variable type hint

  on variables.tf line 2, in variable "test":
   2:   type        = "map(string)"

The legacy variable type hint form, using a quoted string, allows only the
values "string", "list", and "map". To provide a full type expression, remove
the surrounding quotes and give the type expression directly.
```

You can find the reporting here: https://github.com/antonbabenko/pre-commit-terraform/issues/52#issuecomment-505228034

As well as the initial fix including regression tests here: https://github.com/cytopia/docker-terraform-docs/pull/16

An example to reproduce the faulty behaviour (before this PR might be merged) can be found here:
https://github.com/cytopia/docker-terraform-docs/blob/master/tests/0.12/main.tf#L60

### DIFF
--- a/bin/terraform-docs.awk
+++ b/bin/terraform-docs.awk
@@ -61,7 +61,12 @@
       if (type ~ "object") {
         print "  type = \"object\""
       } else {
-        print "  type = \"" $3 "\""
+          # legacy quoted types: "string", "list", and "map"
+          if ($3 ~ /^[[:space:]]*"(.*?)"[[:space:]]*$/) {
+            print "  type = " $3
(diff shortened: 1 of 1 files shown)


aknysh (approved):
thanks @cytopia

PR #158, opened 2019-07-03
Extend build-harness to include modules in another (potentially private) repo

schmidtd:
## What
 
Extend `build-harness` to include modules in another (potentially private) repo.  We do this by adding another directory that is modeled after `build-harness`: it is searched for `modules/*/Makefile*` and are added to the universe of targets just as all the native `build-harness` ones are.

## Why

We have a number of modules and targets that are not at all interesting outside the scope of our project.  So we'd like to use the base `build-harness` just as it is from Cloudposse, but then add our own flavor of modules without having to fork and modify `build-harness` directly.  Then, our builds check out both flavors of `build-harness` (Cloudposse's and ours), and they work together.  This also enables us to easily offer our changes back to the community as we can separate out enhancements we made to the base while keeping our hopelessly proprietary stuff separate.

## Commentary

I'm looking for feedback on the approach here - I'm basically structuring our build in a way that we check out two repos, `build-harness` (from you) and `build-harness-extensions` (private to me), and they are tied together by the `include` in the main `Makefile`.  (We also make use of `BUILD_HARNESS_ARCH` for several things, and I'm also including it here to show how we use it in Terraform's install target to do a little less hard-coding of the package to download.)


arewm:
Does the order in which the `build-harness`/`build-harness-extensions` are included affect the behavior? Should you make a note on the `README` to document the steps for extending the `build-harness`?

schmidtd:
It mostly doesn't matter the order things are included in, because `make` always complains about clashes.  So I'm finding that I can factor out targets into `build-harness-extensions` that otherwise share the entirety of the "parent" `build-harness` and everything is simply merged at execution time.  Definitely agree on the `README` addition; I have it in my local `build-harness-extensions` repo, but it would also belong in `build-harness`.

osterman (approved):
LGTM

@goruha @aknysh any thoughts?

README.md
```diff
@@ -203,7 +203,12 @@ Available targets:
 
 ```
 
+## Extending `build-harness` with targets from another repo
```
osterman: Please update `README.yaml` instead, then rerun `make readme`

osterman (changes requested):
Update `README.yaml`

README.md
schmidtd: Thanks for the nudge in the right direction for documentation updates.  Sorry for the merge noise!

david-schmidt:
Just as an aside, I thought it would be most appropriate to modify `templates/README.md` as the source for the extension documentation since that's where the "Share the Love" section was also sourced - since they're siblings in terms of content, rather than in `README.yaml`.

schmidtd:
Just as an aside, I thought it would be most appropriate to modify `templates/README.md` as the source for the extension documentation since that's where the "Share the Love" section was also sourced - since they're siblings in terms of content, rather than in `README.yaml`.

osterman (approved):

PR #164, opened 2019-08-02
Support validation for 0.12

goruha:
## What
* Disable validation for terraform `0.12`

## Why
* Validation does not allow to skip variables check.


maximmi (approved):

modules/terraform/Makefile
```diff
@@ -24,7 +24,11 @@ terraform/get-modules:
 
 ## Basic terraform sanity check
 terraform/validate:
+ifeq ("12","$(word 2, $(subst ., ,$(TERRAFORM_VERSION)))")
```
osterman, line 27: https://github.com/cloudposse/test-harness/blob/master/test/terraform/validate.bats#L17-L23

We call validate today on 0.12 in our `test-harness`

PR #166, opened 2019-08-26
templates/README: fixed width avatar images using inline html

tamsky:
## What's changing

Contributor images now defined by inline html.

## Why

Image urls' `size` argument is not being honored.

- https://github.com/tamsky.png?size=150 (which redirects to:)
- https://avatars0.githubusercontent.com/u/96978?s=150&v=4

returns a 420px by 420px png.

My guess is the `size`/`s` arg is no longer functional.

There's no other GHFMD feature available to use that can set an image's rendered width/height, besides inline html. 

See [https://github.com/jgm/pandoc/issues/2554#issuecomment-158936360](https://github.com/jgm/pandoc/issues/2554#issuecomment-158936360)

## Example

Changes built (`make readme`), tested and pushed to
- https://github.com/tamsky/terraform-aws-kops-aws-alb-ingress/blob/test/build-harness/README-changes/README.md

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -319,21 +319,29 @@ Check out [our other projects][github], [follow us on twitter][twitter], [apply 
 {{ if has (datasource "config") "contributors" }}
 ### Contributors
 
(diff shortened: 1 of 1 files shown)


osterman:
@tamsky it has never worked with people without avatars, but it still works for people with avatars. The example is one without an avatar.

https://github.com/osterman.png?size=150

tamsky:
> @tamsky it has never worked with people without avatars

I think that was all I was trying to fix.
Thanks for fixing & closing.

Issue #170, opened 2019-09-25
Generate doc failed if "type" present in object definition


eservent:
Hello,
I tried to use /bin/terraform-docs.awk to generate documentation with terraform-docs.

An error occured with this variable from provider aws, resource "aws_ecs_service", property "ordered_placement_strategy" :

```
variable "ordered_placement_strategy" {
  type = list(object({
    type  = string
    field = string
  }))
  description = "Ordered task placement strategy on EC2 instances"
  default = [
    {
      type  = "spread"
      field = "attribute:ecs.availability-zone"
    }
  ]
}
```
Due to presence of "type" in the object definition, and default value, awk transformation failed to retrieve a valid hcl0.11 content.

Transformation result : 
```
variable "ordered_placement_strategy" {
  type = "object"
  type = "string"
  description = "Ordered task placement strategy on EC2 instances"
  default = [
    {
  type = "spread"
}
```

Thank you for the awk hack anyway 👍

osterman:
Closing as this repository is being deprecated. See #393 for details.

PR #172, opened 2019-10-08
Fix terraform-docs with default value with reserved words

eservent:
If default value has reserved words (description, type), awk failed.
From this PR https://github.com/antonbabenko/pre-commit-terraform/pull/66, I tried to fix default value awk transformation


eservent:
#170

maximmi (approved):

osterman:
replaced by #218

PR #177, opened 2019-11-12
Auto readme

osterman:
## what 
* Add workflow to automatically update README 

## why
* Maintenance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ auto-label:
 		echo "$${module%/}: $${module}**"; \
 	done > .github/$@.yml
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -24,6 +24,10 @@ auto-label:
 		echo "$${module%/}: $${module}**"; \
 	done > .github/$@.yml
 
+# For backwards compatibility to how ever other project works that uses build-harness
```
aknysh:
```suggestion
# For backwards compatibility to however other project works that uses build-harness
```

goruha (approved):

PR #211, opened 2020-06-05
fix dispatch template paths

osterman:
## what
* use correct paths

## why
* otherwise templates don't get installed


johncblandii (approved):
LGTM

PR #218, opened 2020-06-15
Fix terraform docs

osterman:
## what
* deprecate old shell script for supporting HCL2
* upgrade packages to support latest `terraform-docs`

## why
* script doesn't very well support all the terraform variable types
* no reason to keep it around now that `terraform-docs` natively supports HCL2


nitrocode (approved):
lgtm

aknysh (approved):

3h4x:
@osterman What about file `terraform-docs.awk`? Is it still used? No reference to it in `buildharness`

osterman:
@3h4x This was also deprecated by #218 and should have all references removed.

3h4x:
> @3h4x This was also deprecated by #218 and should have all references removed.

I will remove it in https://github.com/cloudposse/build-harness/pull/230

PR #225, opened 2020-06-26
Enable pinning to a tag

charlieparkes:
These changes (1) allow you to pin to a tag (and branch) and (2) cleanly switch between branches / tags without cleaning the project directory.

Originally, I went the route of setting the desired tag via a new environment variable, but with the following requirements, I came to this solution.
* most users (presumably) want to use the latest tag on master
* when a change breaks a build, pinning to a tag is desirable
* you should be able to easily test a branch or switch between branches and tags without cleaning


osterman:
This looks nice! I want to test it but I think we should do this.

charlieparkes:
Closing, stale. I'd still like to see this feature added, if it hasn't already.

PR #227, opened 2020-06-27
Support running all targets in docker

osterman:
## what
* Add support for a new `BUILD_HARNESS_DOCKER_ENABLED` flag (disabled by default)
* Run build harness inside of a docker shell

## why
* Many users lack the requisite tools locally to run the build-harness effectively
* Using docker enables us to ship a working combination of tools

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ ENV INSTALL_PATH /usr/local/bin
 WORKDIR /build-harness
 
 RUN make -s bash/lint make/lint
-RUN make -s template/deps aws/install terraform/install readme/deps
+RUN make -s template/deps aws/install terraform/install packages/install/terraform-docs readme/deps
 RUN make -s go/deps-build go/deps-dev
 
 ENTRYPOINT ["/usr/bin/make"]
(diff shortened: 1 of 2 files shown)


Gowiem (approved):
This is clever! Will be interesting to try it out.

nabadger:
@osterman I was wondering if there's any interest in progressing this?

I'm finding having a docker image for build-dependencies is working better than local `make packages/install`, because quite a few packages either don't have a github-release, or they are just src, and require compilation (doesn't really fit into the packages framework). Bundling everything into a docker image is a nice way to get around this (and also turns out to be very useful in CI).

Note, I think this feature would perhaps be more useful if any docker image could be specified
```
  $(BUILD_HARNESS_ORG)/$(BUILD_HARNESS_PROJECT):$(BUILD_HARNESS_DOCKER_TAG) 
```

Although i guess the docker-tag could change between projects importing this...so maybe already achievable (although we tend to have build-harness hosted (forked) on github, but our docker images tend to be private, perhaps on a different platform entirely).

osterman:
So I lost steam on this after I realized there was no elegant way to handle all the environment variables which the build-harness depends on. Short of doing something like...

```
printenv > /dev/shm/env
docker run --env-file=/dev/shm/env ....
```
And that seems really insecure.

We still ship a docker image of the build-harness and you can use that in your pipelines (like we do)

Nuru:
Superseded by #266 and #268

PR #230, opened 2020-07-13
Changes to support README.md to passing super-linter

3h4x:
## what
* Small changes for README.md to be linted

## why
* To conform with MD standards

## references
* https://github.com/cloudposse/.github/pull/1 - `gh actions` are passing for README.md


modules/docs/Makefile
```diff
@@ -6,8 +6,8 @@ docs/deps::
 ## Update `docs/targets.md` from `make help`
 docs/targets.md: docs/deps
 	@( \
-		echo "## Makefile Targets"; \
-		echo '```'; \
+		echo "# Makefile Targets"; \
```
osterman: This is erroneous. The file is included in the `README.md` and should be at the `##` depth. Can we configure the linter to ignore this file? Many linters allow adding an exclusion. 

I guess we could add this file also to `.gitignore` and not check it in.

https://github.com/igorshubovych/markdownlint-cli#ignoring-files

modules/github/Makefile.init
```diff
@@ -1,6 +1,7 @@
 GITHUB_TEMPLATES = \
 	.github/CODEOWNERS \
 	.github/PULL_REQUEST_TEMPLATE.md \
+	.github/workflows/slash-command-dispatch.yml \
```
osterman: This is deprecated and replaced by `chatops.yml`

We should make sure that all references to `slash-command-dispatch.yml` are replaced with `chatops.yml`

osterman (changes requested):
See comments

modules/docs/Makefile
3h4x: Good catch. I didn't notice it's included in README. 
Yep, I'll use the ignore file to mitigate that.

modules/github/Makefile.init
3h4x: Aha, yeah. This line was a mistake. Good catch.

README.md
```diff
@@ -264,7 +244,7 @@ Available targets:
   travis/docker-tag-and-push          Tag & Push according Travis environment variables
 
 ```
-## Extending `build-harness` with targets from another repo
+# Extending `build-harness` with targets from another repo
```
osterman: Why do you think it wants this to be an H1?

osterman: (because the following 2 sections `## Share the Love` and `## References` are not subsections of this.)

docs/extensions.md
```diff
@@ -1,4 +1,4 @@
-## Extending `build-harness` with targets from another repo
+# Extending `build-harness` with targets from another repo
```
osterman: Aha, we should presumably exclude this file then as well from listing. It will be picked up by the README.md lint.

README.md
3h4x: Exactly due to this

`first-line-heading/first-line-h1`

> File:[docs/terraform.md]
> ERROR! Found errors in [markdownlint] linter!
> ERROR:[docs/terraform.md:1 MD041/first-line-heading/first-line-h1 First line in file should be a top level heading >? [Context: "## Requirements"]]

https://github.com/cloudposse/.github/runs/865479515

docs/extensions.md
3h4x: I added such ignore file in this commit:
https://github.com/cloudposse/build-harness/pull/230/commits/31362107d0fdd97964c1f31071f98a76915a863b

modules/docs/Makefile
3h4x: ⬆️ What you provided is actually different repository.
This one is markdownlint used by superlinter
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md041

docs/extensions.md
```diff
@@ -1,4 +1,6 @@
+<!-- markdownlint-disable-file MD041 -->
```
osterman: Nice!

osterman (dismissed):

3h4x:
@osterman I made additional changes as linter still didn't like no `#` at the beginning of `md` file.
https://github.com/cloudposse/build-harness/pull/230/files/d94c7d2773930108a082e18042a8bd826a48537f..28702fe1da5b78ee8a4f3a0232b24ff9a594aded

Now `.github` repo pass linting
Screenshot 2020-07-14 at 14 45 41

**Please reapprove**

osterman (approved):

PR #237, opened 2020-08-14
add dependabot configuration

osterman:
## what
* add dependabot configuration

## why
* automatic module PRs

### DIFF
--- a/templates/.github/dependabot.yml
+++ b/templates/.github/dependabot.yml
@@ -1,0 +1,44 @@
+# Dependabot configuration
+# https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    rebase-strategy: "auto"
+    open-pull-requests-limit: 5
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "github-actions"
+
+  # Maintain dependencies for Terraform modules
+  - package-ecosystem: "terraform"
+    rebase-strategy: "auto"
+    open-pull-requests-limit: 5
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "terraform"
+
+  # Maintain dependencies for Terraform Submodules
+  - package-ecosystem: "terraform"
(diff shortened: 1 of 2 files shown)


templates/.github/dependabot.yml
```diff
@@ -0,0 +1,44 @@
...
+    schedule:
+      interval: "daily"
+    labels:
+      - "terraform-submodules"
+
+  # Maintain dependencies for Terraform Examples
+  - package-ecosystem: "terraform"
+    rebase-strategy: "auto"
+    open-pull-requests-limit: 5
+    directory: "/examples/complete/"
```
nitrocode, line 40: why not all examples ?

lorengordon:
Heads up that dependabot upstream _still_ [does not support HCL2](https://github.com/dependabot/dependabot-core/pull/1388). There are a couple options that function as a separate github action, though...

* https://github.com/patrickjahns/dependabot-terraform-action
* https://github.com/plus3it/dependabot-terraform-action (our fork, mostly shrinking the dependencies, iirc)

Something like this is what we use:
```
name: Update terraform dependencies

on:
  schedule:
    - cron:  '0 6 * * *'

jobs:
  dependabot-terraform:
    runs-on: ubuntu-latest
    steps:
      - name: update terraform dependencies
        uses: plus3it/dependabot-terraform-action@p3-implementation
        with:
          github_dependency_token: ${{ secrets.GH_READONLY_TOKEN }}
```

templates/.github/dependabot.yml
osterman: Unfortunately, I don't think there's a way to tell it to go recursively. I would need to know what all examples are upfront or generate the file (which is out-of-scope for this PR).

osterman:
Ah... I remember someone saying that on office hours. I could have sworn it was working for us, but I looked back and the PR history and don't see any coming in anymore for terraform, so I guess that's why.

@lorengordon does this action handle the rebasing?

lorengordon:
> @lorengordon does this action handle the rebasing?

I don't believe so, no... That's part of the Dependabot SaaS offering, not the packaged container. Maybe there's another github action that could do that? Or maybe mergify?

Nuru:
Looks like Renovate is a better option. We can revisit/reopen this later if we change our minds.

PR #242, opened 2020-08-15
Updates to auto-release.yml files to include PR body + use minor > patch as default

Gowiem:
## what

* Updates change-template to include the PR body 
* Updates to bump the minor as the default instead of bumping patch 

## why
* Our PR bodies typically have good information in them due to our PR body template being simple and straightforward. It's good to include that in our release notes as most folks likely won't check the PR itself. 
* We do minor releases much more frequently than patch so using minor as our default makes sense to save us time in regards to merging without a label.

### DIFF
--- a/.github/auto-release.yml
+++ b/.github/auto-release.yml
@@ -29,7 +29,10 @@ categories:
       - 'bug'
       - 'hotfix'
 
-change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
+change-template: |
+  ## $TITLE @$AUTHOR (#$NUMBER)
+
+  $BODY
 
 template: |
   $CHANGES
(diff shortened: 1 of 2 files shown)


.github/auto-release.yml
```diff
@@ -29,7 +29,10 @@ categories:
       - 'bug'
       - 'hotfix'
 
-change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
+change-template: |
+  ## $TITLE @$AUTHOR (#$NUMBER)
```
osterman, line 33: So I considered doing this, but our descriptions have `##` in them usually. 

image

See #241

osterman: https://github.com/cloudposse/build-harness/releases/tag/0.40.0

PR #262, opened 2020-12-07
Add support for Terraform 0.14

Nuru:
## what

- Add target that starts the build-harness Docker image and lets you run `make` or `bash` from within the container
- Add target to update README for Terraform modules to switch to registry format
- Add target to bump modules with minimum core version of '0.12.x' to '>= 0.12.26'
- Replace constraints on upper bounds of versions with constraints only on lower bounds. `~> X` converted to `>= X`, explicit upper limits like `< 0.14` removed completely.
- Add target to rewrite `versions.tf` to add an explicit source attribute for each provider
- Add group target for all one-time Terraform updates
- Update `github/init` templates to include `validate-codeowners` workflow and add Terraform-only `.github/*` files

## why

- Allow users to use the scripts and tools without having to install binaries on their host computers
- Switch README to preferred registry format
- Explicit provider source addresses were introduced with Terraform v0.13, so the full provider requirements syntax is not supported by Terraform v0.12, but v0.12.26 will accept but ignore the source argument in a required_providers block
- Terraform v0.13 introduces new syntax in the `required_providers` nested block inside the terraform configuration block. We want to make modules compatible with Terraform v0.14, which drops support for the v0.12 format.
- Make it easier to apply all updates at once
- Invalid CODEOWNERS becomes silently ignored, removing all protection, so it is critical to ensure it is valid

## ToDo

- Enhance `github/init` to allow for multiple versions of the same file, installing correct one based on repo nature
- Configure non-Terraform version of Codeowners
- Restore file existence check to verify-codeowners.

## references

- See example transformations [in this PR](https://github.com/cloudposse/terraform-aws-datadog-integration/pull/18/files#)
- [v0.12-Compatible Provider Requirements](https://www.terraform.io/docs/configuration/provider-requirements.html#v0-12-compatible-provider-requirements)
- [Explicit Provider Source Locations](https://www.terraform.io/upgrade-guides/0-13.html#explicit-provider-source-locations)
- Supersedes and closes #264, closes #247, closes #255, closes #257, closes #259, closes #263


modules/terraform/Makefile
```diff
@@ -46,12 +46,19 @@ terraform/upgrade-modules:
 
 ## Rewrite the *.tf files to use registry notation for modules sources
 terraform/rewrite-module-source:
-	@sed -i -E 's,"git::https://github.com/(.*?)/terraform-([^-]*?)-(.*?).git\?ref=tags/(.*?)","\1/\3/\2"\n  version     = "\4",g' $$(find . -type f -name '*.tf')
+	@sed -i -E 's,\s*source\s+=\s+"git::https://github.com/(.*?)/terraform-([^-]*?)-(.*?).git\?ref=(tags/)?(.*?)",  source = "\1/\3/\2"\n  version     = "\5",g' $$(find . -type f -not -name context.tf -name '*.tf')
 	@$(TERRAFORM) fmt .
 	@$(TERRAFORM) fmt examples/complete
 
+terraform/rewrite-readme-source:
+	@sed -i -E 's,^(\s*)source\s+=\s+"git::https://github.com/(.*?)/terraform-([^-]*?)-(.*?).git\?ref=(tags/)?master",\1source = "\2/\4/\2"\n\1# Cloud Posse recommends pinning module to a specific version\n\1# version     = "x.x.x",g' README.yaml
```
maximmi:
```suggestion
	@sed -i -E 's,^(\s*)source\s+=\s+"git::https://github.com/(.*?)/terraform-([^-]*?)-(.*?).git\?ref=(tags/)?master",\1source = "\2/\4/\3"\n\1# Cloud Posse recommends pinning module to a specific version\n\1# version     = "x.x.x",g' README.yaml
```

maximmi (changes requested):

maximmi (dismissed):

modules/terraform/Makefile
```diff
@@ -46,12 +46,36 @@ terraform/upgrade-modules:
...
 	@$(TERRAFORM) fmt examples/complete
+
+## Rewrite versions.tf to bump modules with minimum core version of '0.12.x' to '>= 0.12.26'
+terraform/bump-tf-12-min-version:
+	@sed -i -E 's,required_version\s*\=\s*\"(\~>|>\=)\s?(0\.12\.0)\",required_version = ">= 0.12.26",g' $$(find . -type f -name 'versions.tf')
+	@$(TERRAFORM) fmt .
+	@$(TERRAFORM) fmt examples/complete
+
+## Rewrite versions.tf to update existing configuration to add an explicit source attribute for each provider
+terraform/rewrite-required-providers:
```
osterman: How will we handle sub modules? Like we have in Datadog monitors and the tfc module?

Nuru: Submodules do not specify provider constraints. (At least Cloud Posse submodules don’t)

maximmi: and they are using relative paths for modules source, as well as I can see so far

modules/terraform/Makefile
```diff
@@ -46,12 +46,36 @@ terraform/upgrade-modules:
...
+
 ## Rewrite versions.tf to remove upper bound for terraform core version constraint (like this ">= 0.12.0, < 0.14.0")
 terraform/remove-upper-bound:
 	@sed -i -E 's,required_version\s*\=\s*\"(.*?)(\,\s*<.*)",required_version = "\1",g' $$(find . -type f -name 'versions.tf')
 	@$(TERRAFORM) fmt .
 	@$(TERRAFORM) fmt examples/complete
+
+## Rewrite versions.tf to bump modules with minimum core version of '0.12.x' to '>= 0.12.26'
+terraform/bump-tf-12-min-version:
+	@sed -i -E 's,required_version\s*\=\s*\"(\~>|>\=)\s?(0\.12\.0)\",required_version = ">= 0.12.26",g' $$(find . -type f -name 'versions.tf')
```
Nuru: Configured version might not be 0.12.0. Could be 0.12.20 or just 0.12

modules/terraform/Makefile
```diff
@@ -46,12 +46,34 @@ terraform/upgrade-modules:
...
+	@$(TERRAFORM) fmt .
+	@$(TERRAFORM) fmt examples/complete
+
+## Rewrite versions.tf to update existing configuration to add an explicit source attribute for each provider
+terraform/rewrite-required-providers:
+	@KEEP_VERSION=$(grep -Eo 'required_version\s*=\s*"(.*)"' versions.tf | cut -d '"' -f2)
+	@terraform-0.13 0.13upgrade -yes ./
```
Nuru: Does this work if the module has already been upgraded? If not, need to check version first.

Nuru: OK, it does work

maximmi (dismissed):

modules/packages/Makefile
```diff
@@ -13,18 +14,23 @@ packages/reinstall: packages/delete packages/install
...
-packages/install/%: packages/install
-	@if [ ! -x $(INSTALL_PATH)/$(subst packages/install/,,$@) ]; then \
-		$(MAKE) -C $(PACKAGES_PATH)/install $(subst packages/install/,,$@); \
+packages/install/%:
+	@binary="$*"; \
+	if [[ -x "$(INSTALL_PATH)/$$binary" ]]; then \
+  		echo "* Package $$binary already installed"; \
+	elif [[ "$(PACKAGES_PREFER_HOST)" == "true" ]] && installed=$$(command -v $* 2>/dev/null); then \
+			echo Using "$*" from "$$installed" ; \
+			cp -p "$$installed" "$*" ; \
```
osterman: `ln`?

templates/terraform/.github/mergify.yml
```diff
@@ -0,0 +1,51 @@
...
+  - "merged"
+  actions:
+    delete_head_branch: {}
+
+- name: "ask to resolve conflict"
+  conditions:
+  - "conflict"
+  actions:
+    comment:
+      message: "This pull request is now in conflicts. Could you fix it @{{author}}? 🙏"
```
osterman:
```suggestion
      message: "This pull request is now in conflict. Could you fix it @{{author}}? 🙏"
```

templates/terraform/.github/workflows/auto-context.yml
```diff
@@ -0,0 +1,54 @@
...
+        if [[ -f context.tf ]]; then
+          echo "Discovered existing context.tf! Fetching most recent version to see if there is an update."
+          curl -o context.tf -fsSL https://raw.githubusercontent.com/cloudposse/terraform-null-label/master/exports/context.tf
+          if git diff --no-patch --exit-code context.tf; then
+            echo "No changes detected! Exiting the job..."
+          else
+            echo "context.tf file was updated. Need to rebuild README.md."
+            make init
+            make readme/build
+            echo "::set-output name=create_pull_request=true"
```
osterman, line 30: Technically unnecessary because `create-pull-request` will only open a PR if there were changes.

templates/Makefile.build-harness
```diff
@@ -16,4 +23,28 @@ init::
...
+	@docker pull $(BUILD_HARNESS_DOCKER_IMAGE):$(BUILD_HARNESS_DOCKER_TAG)
+	@exit 0
+
+DEFAULT_DOCKER_ENVS := AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN TERM AWS_PROFILE AWS_REGION \
+ 	AWS_DEFAULT_PROFILE AWS_DEFAULT_REGION
+EXTRA_DOCKER_ENVS ?= AWS_CONFIG_FILE AWS_SHARED_CREDENTIALS_FILE
+MOUNT_HOME ?= $(shell [ -d "$$HOME" ] && printf -- "-e HOME -v \"%s\":\"%s\"" "$$HOME" "$$HOME")
+DOCKER_ENVS ?= $(DEFAULT_DOCKER_ENVS) $(EXTRA_DOCKER_ENVS)
+
+shell builder:
```
osterman, line 45:
```suggestion
## Start a shell inside of the `build-harness` docker container
shell builder:
```

osterman (dismissed):
Fix auto-release config

modules/packages/Makefile
Nuru: I originally wrote `ln` but then realized that the `vendor` directory will probably be on the host and the `bin` directory could be in the Docker image, so `ln` could break things for one or the other.

templates/terraform/.github/workflows/auto-context.yml
Nuru: `create-pull-request` will still do a lot of work to determine that no changes were needed, and since we will be running over 100 of these in parallel, and the normal case will be no changes, I want to short-circuit as much as possible.

modules/packages/Makefile
Nuru: Actually, we can do without `cp` or `ln`, since the executable is already on the `PATH`

modules/github/Makefile.init
```diff
@@ -6,16 +6,33 @@ GITHUB_TEMPLATES = \
...
 	cp $(BUILD_HARNESS_PATH)/templates/$@ $@
 	git ls-files --error-unmatch $@ 2>/dev/null || git add $@
 
-github/init: $(GITHUB_TEMPLATES)
+$(GITHUB_TERRAFORM_TEMPLATES): $(addprefix $(BUILD_HARNESS_PATH)/templates/terraform/, $(GITHUB_TERRAFORM_TEMPLATES))
+	mkdir -p $(dir $@)
+	cp $(BUILD_HARNESS_PATH)/templates/terraform/$@ $@
+	git ls-files --error-unmatch $@ 2>/dev/null || git add $@
+
+context:
```
osterman: `context` is an overly broad target name. These target names are global.

Nuru: I forgot these are massively included. Changed to `github/init/context.tf`

maximmi (dismissed):

osterman (approved):

PR #264, opened 2020-12-07
terraform and tools to work with it are added to image

maximmi:
## what
* `terraform` installed (versions 0.11-0.14)
* `terraform-config-inspect` installed
* `vert` installed

## why
* needed for some Makefiles

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,11 @@ RUN apk update && \
 
 RUN curl -sSL https://apk.cloudposse.com/install.sh | bash
 
+## Install terraform-config-inspect (required for bats tests)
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -23,6 +23,11 @@ RUN apk update && \
 
 RUN curl -sSL https://apk.cloudposse.com/install.sh | bash
 
+## Install terraform-config-inspect (required for bats tests)
+ENV GO111MODULE="on"
+RUN go get github.com/hashicorp/terraform-config-inspect && \
```
osterman, line 28: We can soon replace this with the package. https://github.com/cloudposse/packages/pull/748

PR #266, opened 2020-12-12
Add commands to help create pull requests

Nuru:
## what
- Create a slim Docker image to run commands that help with PRs
- Add `pr/prepare` and `tf14-upgrade` commands
- Update/fix workflows

## why
- Help automate the task of updating all our Terraform modules to support Terraform 0.14

### DIFF
--- a/modules/terraform/Makefile
+++ b/modules/terraform/Makefile
@@ -40,6 +40,20 @@ else
 	[ -z "$$FAIL" ] || (echo "Terraform configuration needs linting. Run '$(TERRAFORM) fmt'"; echo $$FAIL; exit 1)
 endif
 
+terraform/fmt:
+ifeq ($(wildcard *.tf),)
+	@echo "* $@: No terraform files detected"
+else
+	@{ command -v terraform-0.13 && command -v terraform-0.14; } >/dev/null || { echo "* $@ requires binaries named terraform-0.13 and terraform-0.14" && false; }
+	@ # vert exits non-zero if any of the versions are not acceptable, so `|| [[ -n "$VERSION" ]]` for a real error check
+	@TF12=0.12.29; \
+	TF13=$$(terraform-0.13 version --json | jq -r .terraform_version); \
+	TF14=$$(terraform-0.14 version --json | jq -r .terraform_version); \
+	VERSION=$$(vert -s "$$(terraform-config-inspect --json . | jq -r '.required_core[]')" "$$TF12" "$$TF13" "$$TF14" | head -1) || [[ -n "$$VERSION" ]]; \
+	VERSION=$${VERSION:0:4}; \
+	terraform-$${VERSION} fmt -recursive .
+endif
+
 ## Upgrade all terraform module sources
 ## This target has not been upgraded to handle registry format
 terraform/upgrade-modules:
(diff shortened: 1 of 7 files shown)


modules/terraform/Makefile
```diff
@@ -40,6 +40,20 @@ else
...
+	@echo "* $@: No terraform files detected"
+else
+	@{ command -v terraform-0.13 && command -v terraform-0.14; } >/dev/null || { echo "* $@ requires binaries named terraform-0.13 and terraform-0.14" && false; }
+	@ # vert exits non-zero if any of the versions are not acceptable, so `|| [[ -n "$VERSION" ]]` for a real error check
+	@TF12=0.12.29; \
+	TF13=$$(terraform-0.13 version --json | jq -r .terraform_version); \
+	TF14=$$(terraform-0.14 version --json | jq -r .terraform_version); \
+	VERSION=$$(vert -s "$$(terraform-config-inspect --json . | jq -r '.required_core[]')" "$$TF12" "$$TF13" "$$TF14" | head -1) || [[ -n "$$VERSION" ]]; \
+	VERSION=$${VERSION:0:4}; \
+	terraform-$${VERSION} fmt -recursive .
```
osterman, line 54: We have to be backward compatible on this one. 99% of terraform users use our binary name format, nor will they have or can we require they have vert. Build harness is not used exclusively by cloud Posse CI.

Also, to support new versions of terraform, we have very few places we must update, and most of those places do not base the logic on the major version. I am not eager to start down this path.

templates/terraform/.github/workflows/auto-context.yml
```diff
@@ -1,55 +1,55 @@
...
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      run: |
+        if [[ -f context.tf ]]; then
+          echo "Discovered existing context.tf! Fetching most recent version to see if there is an update."
+          curl -o context.tf -fsSL https://raw.githubusercontent.com/cloudposse/terraform-null-label/master/exports/context.tf
+          if git diff --no-patch --exit-code context.tf; then
+            echo "No changes detected! Exiting the job..."
           else
-            echo "This module has not yet been updated to support the context.tf pattern! Please update in order to support automatic updates."
+            echo "context.tf file has changed. Update examples and rebuild README.md."
```
osterman, line 26: We have modules that do not use null label. I don't know how I feel about this because this is just updating one file and moving to context pattern is a lot more work. What is the benefit from this?

osterman (changes requested):

modules/terraform/Makefile
Nuru: This is a new target, so we do not need to be backwards compatible. It is meant to be run from within the Build Harness Docker image, and fails with a nice error message if the binaries are not present. 

We can talk about packaging this up into something DRY, but the bulk of it is copy/pasted from the `test-command` workflow.

templates/terraform/.github/workflows/auto-context.yml
Nuru: This is the automatic update of `context.tf` you approved a week ago. All I did was revert whitespace changes from #265 and restore the cron job quotes.

osterman: I missed that we introduced this 2nd type of PR for context. Not keen for it and it's potential to generate more PRs that we cannot look into and PRs that won't help in any implementation

Nuru: View the file, not the diff. All this does for modules that do not have `context.tf` already is output a message in the action log, not open a PR.

templates/terraform/.github/workflows/auto-context.yml
```diff
@@ -1,55 +1,55 @@
...
+            make init
+            make github/init/context.tf
+            make readme/build
+            echo "::set-output name=create_pull_request=true"
```
osterman, line 30:
```suggestion
            echo "::set-output name=create_pull_request=false"
```

osterman (approved):

templates/terraform/.github/workflows/auto-context.yml
osterman: disregard

PR #272, opened 2021-01-08
Adds auto-init capability using the bootstrap include directive

lorengordon:
## what

These changes will automatically run the `make init` workflow, and keep the clone up-to-date with the target branch/tag.

## why

1. People on my team kept forgetting to run `make init`, and it proved sufficiently frustrating (and often) that I figured I'd try to automate it.

2. This makes it possible for the build-harness to update itself automatically. The user's Makefile can include `BUILD_HARNESS_BRANCH=<tag>` and `BUILD_HARNESS_AUTO_INIT=true`, and they can have a mechanism that updates that tag value, and the auto-init logic will ensure that the checkout is updated automatically.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,6 @@ RUN update-alternatives --set terraform /usr/share/terraform/$DEFAULT_TERRAFORM_
   cp -p /usr/share/terraform/$DEFAULT_TERRAFORM_VERSION/bin/terraform /build-harness/vendor/terraform
 
 COPY ./ /build-harness/
-RUN touch /.dockerenv
 
 ENV INSTALL_PATH /usr/local/bin
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,6 +53,7 @@ RUN update-alternatives --set terraform /usr/share/terraform/$DEFAULT_TERRAFORM_
   cp -p /usr/share/terraform/$DEFAULT_TERRAFORM_VERSION/bin/terraform /build-harness/vendor/terraform
 
 COPY ./ /build-harness/
+RUN touch /.dockerenv
 
 ENV INSTALL_PATH /usr/local/bin
 
--- a/templates/Makefile.build-harness
+++ b/templates/Makefile.build-harness
@@ -37,6 +37,10 @@ if [[ \
 ]]; then \
 	echo "[.build-harness]: In $(BUILD_HARNESS_PROJECT) docker container, skipping auto-init" ;\
 elif [[ \
+	grep -q docker /proc/1/cgroup 2>/dev/null \
+]]; then \
+	echo "[.build-harness]: In unknown docker container, skipping auto-init" ;\
+elif [[ \
 	"$(BUILD_HARNESS_PATH)" != "$(BUILD_HARNESS_PATH_LOCAL)" && \
 	-f "$(BUILD_HARNESS_PATH)/Makefile" \
 ]]; then \
--- a/templates/Makefile.build-harness
+++ b/templates/Makefile.build-harness
@@ -29,6 +29,10 @@ endef
 # checkout does not match BUILD_HARNESS_BRANCH
 define harness_auto_init
 if [[ \
+	-f "/.dockerenv" && -f "/$(BUILD_HARNESS_PROJECT)/Makefile" \
+]]; then \
+	: ;\
+elif [[ \
 	"$(BUILD_HARNESS_PATH)" != "$(BUILD_HARNESS_PATH_LOCAL)" && \
 	-f "$(BUILD_HARNESS_PATH)/Makefile" \
 ]]; then \
--- a/templates/Makefile.build-harness
+++ b/templates/Makefile.build-harness
@@ -4,21 +4,52 @@
 #
 
 export SHELL = /bin/bash
-export BUILD_HARNESS_PATH ?= $(shell until [ -d "$(BUILD_HARNESS_PROJECT)" ] || [ "`pwd`" == '/' ]; do cd ..; done; pwd)/$(BUILD_HARNESS_PROJECT)
 export BUILD_HARNESS_ORG ?= cloudposse
 export BUILD_HARNESS_PROJECT ?= build-harness
 export BUILD_HARNESS_DOCKER_IMAGE ?= $(BUILD_HARNESS_ORG)/$(BUILD_HARNESS_PROJECT)
 export BUILD_HARNESS_BRANCH ?= master
+export BUILD_HARNESS_CLONE_URL ?= https://github.com/$(BUILD_HARNESS_ORG)/$(BUILD_HARNESS_PROJECT).git
 # It is kind of expensive to figure out the Docker SHA tag, so we just define the command here, and only call it when needed
 export BUILD_HARNESS_DOCKER_SHA_TAG_CMD := git -C "$(BUILD_HARNESS_PATH)" log -n 1 --format=sha-%h 2>/dev/null || echo latest
 
--include $(BUILD_HARNESS_PATH)/Makefile
+# Resolves BUILD_HARNESS_PATH to BUILD_HARNESS_PATH_LOCAL when BUILD_HARNESS_PATH does not exist
+BUILD_HARNESS_PATH ?= $(shell until [ -d "$(BUILD_HARNESS_PROJECT)" ] || [ "`pwd`" == '/' ]; do cd ..; done; pwd)/$(BUILD_HARNESS_PROJECT)
+BUILD_HARNESS_PATH_LOCAL := $(PWD)/$(BUILD_HARNESS_PROJECT)
+export BUILD_HARNESS_PATH := $(or $(wildcard $(BUILD_HARNESS_PATH)),$(BUILD_HARNESS_PATH_LOCAL))
+
+# Macro to clone/install BUILD_HARNESS_PROJECT
+define harness_install
+curl --retry 5 --fail --silent --retry-delay 1 \
+	https://raw.githubusercontent.com/$(BUILD_HARNESS_ORG)/$(BUILD_HARNESS_PROJECT)/$(BUILD_HARNESS_BRANCH)/bin/install.sh | \
+	bash -s "$(BUILD_HARNESS_ORG)" "$(BUILD_HARNESS_PROJECT)" "$(BUILD_HARNESS_BRANCH)"
+endef
+
+# Macro to auto-init the BUILD_HARNESS_PROJECT with the `include` directive
+# Tests if BUILD_HARNESS_PROJECT does not yet exist, or if it does exist but the
+# checkout does not match BUILD_HARNESS_BRANCH
+define harness_auto_init
+if [[ \
+	"$(BUILD_HARNESS_PATH)" != "$(BUILD_HARNESS_PATH_LOCAL)" && \
+	-f "$(BUILD_HARNESS_PATH)/Makefile" \
+]]; then \
+	: ;\
+elif [[ \
+	"$(BUILD_HARNESS_PATH)" == "$(BUILD_HARNESS_PATH_LOCAL)" && \
+	-f "$(BUILD_HARNESS_PATH)/Makefile" && \
+	"$$(git -C '$(BUILD_HARNESS_PATH_LOCAL)' ls-remote '$(BUILD_HARNESS_CLONE_URL)' '$(BUILD_HARNESS_BRANCH)' | cut -f1)" == "$$(git -C '$(BUILD_HARNESS_PATH_LOCAL)' rev-parse HEAD)" \
+]]; then \
+	: ;\
+else \
+	$(harness_install) ;\
+fi
+endef
+
+-include $(shell $(harness_auto_init) >&2 ; echo $(BUILD_HARNESS_PATH)/Makefile)
 
 .PHONY : init
 ## Init build-harness
 init::
-	@curl --retry 5 --fail --silent --retry-delay 1 https://raw.githubusercontent.com/$(BUILD_HARNESS_ORG)/$(BUILD_HARNESS_PROJECT)/$(BUILD_HARNESS_BRANCH)/bin/install.sh | \
-		bash -s "$(BUILD_HARNESS_ORG)" "$(BUILD_HARNESS_PROJECT)" "$(BUILD_HARNESS_BRANCH)"
+	@ $(harness_install)
 
 .PHONY : clean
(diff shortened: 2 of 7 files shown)


templates/Makefile.build-harness
```diff
@@ -4,21 +4,52 @@
...
 export BUILD_HARNESS_DOCKER_SHA_TAG_CMD := git -C "$(BUILD_HARNESS_PATH)" log -n 1 --format=sha-%h 2>/dev/null || echo latest
 
--include $(BUILD_HARNESS_PATH)/Makefile
+# Resolves BUILD_HARNESS_PATH to BUILD_HARNESS_PATH_LOCAL when BUILD_HARNESS_PATH does not exist
+BUILD_HARNESS_PATH ?= $(shell until [ -d "$(BUILD_HARNESS_PROJECT)" ] || [ "`pwd`" == '/' ]; do cd ..; done; pwd)/$(BUILD_HARNESS_PROJECT)
+BUILD_HARNESS_PATH_LOCAL := $(PWD)/$(BUILD_HARNESS_PROJECT)
+export BUILD_HARNESS_PATH := $(or $(wildcard $(BUILD_HARNESS_PATH)),$(BUILD_HARNESS_PATH_LOCAL))
+
+# Macro to clone/install BUILD_HARNESS_PROJECT
+define harness_install
```
lorengordon, line 25: If desired, we could also inline the `install.sh` script to eliminate the network call:

```suggestion
define harness_install
if [[ -d "$(BUILD_HARNESS_PATH_LOCAL)" ]]; then \
  echo "Removing existing $(BUILD_HARNESS_PATH_LOCAL)" &&\
  rm -rf "$(BUILD_HARNESS_PATH_LOCAL)" ;\
fi && \
echo "Cloning $(BUILD_HARNESS_CLONE_URL)?ref=$(BUILD_HARNESS_BRANCH)..." &&\
git clone -c advice.detachedHead=false --depth=1 -b "$(BUILD_HARNESS_BRANCH)" "$(BUILD_HARNESS_CLONE_URL)"
endef
```

jamengual (commented):
This looks useful

Nuru (changes requested):
This may be OK as is, but `build-harness` is used in multiple ways by multiple automated and manual processes, and so this needs be very carefully evaluated. In particular, certain GitHub Action workflows purposefully avoid calling `make init` because they do not want to use the latest version but rather use the version baked into the Docker image.

Please give more details in the initial PR comment about what problem this is solving and how it is solving it. If all this is solving is that `make xyz` fails because you forgot to run `make init` first, I would rather not risk it.

Also, longer term, I see us migrating more and more `make` targets to run inside the `build-harness` docker image rather than on the host, and this seems to complexify that migration, so I want to weigh that against the purported benefits.

lorengordon:
@Nuru I don't really have more details than I've already explained. Boils down to it be confusing to have to run `make init` in some scenarios and not in others. That logic can be built in, instead, which is what I'm attempting in this PR. But perhaps ya'll just have really forgiving engineers. :) I envy you. If you don't want to accept this, that's certainly fine.

I added a condition to skip the auto-init logic if it's in the build-harness docker container. Sometimes you _do_ want to `init` when in a container, if for some reason it's a _different_ container (ran into this with CodeBuild).

Nuru:
> Boils down to it be confusing to have to run make init in some scenarios and not in others.

@lorengordon I still have not found time to review this in depth, will hopefully get to it in the next few days. Short of that, how about we just output a clear error message that you need to run `make init` if you haven't already? Or just tell your engineers to always run `make init` because there's not much harm in running it every time?

lorengordon:
> just tell your engineers to always run `make init` because there's not much harm in running it every time

Yeah, well, that's what got me here in the first place. 😂 I'm not trying to make anything harder on you. Really if this isn't a direction you want to go, I totally get it. It was just helpful to me, and folks in cloudposse slack seemed interested, so I thought I'd share. We manage our own project internally anyway, so we'll have this even if you choose not to use it.

> how about we just output a clear error message that you need to run `make init` if you haven't already

I think this would end up being 90% the same logic as in the PR now. 😁 Need to detect the conditions where init has or has not occurred, which is the macro `harness_auto_init`. Would probably put that in an `$(if ...` of some kind, with one of the branches handling the error case with `$(error ...`.

lorengordon:
@Nuru I was thinking about this some more, and realized it was pretty straightforward to add a toggle to disable this functionality by default. That way there is no impact to existing projects. I also added some informational output to the auto-init to provide the user with feedback on the auto-init status.

I also thought of another use case, which might be more compelling for you. I updated the PR description, but I'll paste it here also so you see it in your notifications:

> This makes it possible for the build-harness to update itself automatically. The user's Makefile can include `BUILD_HARNESS_BRANCH=<tag>` and `BUILD_HARNESS_AUTO_INIT=true`, and when they update that tag value in their project, the auto-init logic will ensure that the checkout is updated automatically also.

lorengordon:
Some examples of it working...

Using the defaults from this PR with an external clone:
```
$ make help
Available targets:

  aws/install                         Install aws cli bundle
```

Enabling auto-init but using an external clone of build-harness (no auto-init):
```
$ BUILD_HARNESS_AUTO_INIT=true make help
[.build-harness]: Using external /home/loren/git/repos/build-harness, skipping auto-init
Available targets:

  aws/install                         Install aws cli bundle
```

Enabling auto-init but using the build-harness container (no auto-init):
```
$ docker run --rm -ti -v "$PWD:/workdir" -e BUILD_HARNESS_AUTO_INIT=true -w /workdir cloudposse/build-harness help
[.build-harness]: In build-harness docker container, skipping auto-init
Available targets:

[.build-harness]: In build-harness docker container, skipping auto-init  # <--- this is fixed in this PR, by the `export BUILD_HARNESS_AUTO_INIT=false` in the root Makefile
  aws/install                         Install aws cli bundle
```

Enabling auto-init in a new project without an external clone (it should auto-init):
```
$ BUILD_HARNESS_AUTO_INIT=true BUILD_HARNESS_ORG=lorengordon BUILD_HARNESS_BRANCH=auto-init make help
Cloning https://github.com/lorengordon/build-harness.git#auto-init...
Cloning into 'build-harness'...
remote: Enumerating objects: 143, done.
remote: Counting objects: 100% (143/143), done.
remote: Compressing objects: 100% (117/117), done.
remote: Total 143 (delta 8), reused 72 (delta 4), pack-reused 0
Receiving objects: 100% (143/143), 85.86 KiB | 1.43 MiB/s, done.
Resolving deltas: 100% (8/8), done.
Available targets:

  aws/install                         Install aws cli bundle
```

Enabling auto-init with an up-to-date checkout (no auto-init):
```
$ BUILD_HARNESS_AUTO_INIT=true BUILD_HARNESS_ORG=lorengordon BUILD_HARNESS_BRANCH=auto-init make help
[.build-harness]: Clone of build-harness is up-to-date, skipping auto-init
Available targets:

  aws/install                         Install aws cli bundle
```

Enabling auto-init in that same project with a different org/branch (cloudposse/master):
```
$ BUILD_HARNESS_AUTO_INIT=true make help
Removing existing build-harness
Cloning https://github.com/cloudposse/build-harness.git#master...
Cloning into 'build-harness'...
remote: Enumerating objects: 143, done.
remote: Counting objects: 100% (143/143), done.
remote: Compressing objects: 100% (118/118), done.
remote: Total 143 (delta 7), reused 71 (delta 3), pack-reused 0
Receiving objects: 100% (143/143), 85.57 KiB | 2.09 MiB/s, done.
Resolving deltas: 100% (7/7), done.
Available targets:

[.build-harness]: Clone of build-harness is up-to-date, skipping auto-init  # <--- this is fixed in this PR, by the `export BUILD_HARNESS_AUTO_INIT=false` in the root Makefile

  aws/install                         Install aws cli bundle
```

Dockerfile
```diff
@@ -53,6 +53,7 @@ RUN update-alternatives --set terraform /usr/share/terraform/$DEFAULT_TERRAFORM_
   cp -p /usr/share/terraform/$DEFAULT_TERRAFORM_VERSION/bin/terraform /build-harness/vendor/terraform
 
 COPY ./ /build-harness/
+RUN touch /.dockerenv
```
Nuru: This is kind of  hack, and not my kind of hack. :-)
```suggestion
```

templates/Makefile.build-harness
Nuru: The reason we have the network call with branch `master` is so that we can update the script without updating the docker image. In other words, we want the network call.

templates/Makefile.build-harness
```diff
@@ -4,21 +4,60 @@
...
+endef
+
+# Macro to auto-init the BUILD_HARNESS_PROJECT with the `include` directive
+# Tests if BUILD_HARNESS_PROJECT does not yet exist, or if it does exist but the
+# checkout does not match BUILD_HARNESS_BRANCH
+define harness_auto_init
+if [[ \
+	-f "/.dockerenv" && -f "/$(BUILD_HARNESS_PROJECT)/Makefile" \
+]]; then \
+	echo "[.build-harness]: In $(BUILD_HARNESS_PROJECT) docker container, skipping auto-init" ;\
```
Nuru:
```suggestion
	-f /build-harness/Makefile || -f "/$(BUILD_HARNESS_PROJECT)/Makefile"  \
]]; then \
	echo "[.build-harness]: In $(BUILD_HARNESS_PROJECT) docker container, skipping auto-init" ;\
elif [[ \
	grep -q docker /proc/1/cgroup 2>/dev/null \
]]; then \
	echo "[.build-harness]: In unknown docker container, skipping auto-init" ;\
```

Nuru (changes requested):
I worry about unintended consequences, but this is good work, so I am not _too_ worried.

Please make the requested changes and I will approve it. Thank you for the contribution.

Dockerfile
lorengordon: Do you have another suggestion to improve detection you are operating within the docker container? I searched but found even worse hacks, or hacks that were platform-dependent.

lorengordon: Oh, I see, you want to use `/proc/1/cgroup`. Alright, that works as long as you are actually using docker and not another container runtime. If that's good enough for you, no problem.

templates/Makefile.build-harness
lorengordon: This does not feel quite right... If using a _different_ docker container, you almost certainly *do* want to auto-init, especially if the user has set `BUILD_HARNESS_AUTO_INIT=true` which is required right now, because that container will not have the build-harness.

lorengordon: For example, using the `python:3` container:

```
$ docker run --rm -ti -v "$PWD:/workdir" -e BUILD_HARNESS_AUTO_INIT=true -w /workdir --entrypoint make python:3 help
Cloning https://github.com/cloudposse/build-harness.git#master...
Cloning into 'build-harness'...
remote: Enumerating objects: 143, done.
remote: Counting objects: 100% (143/143), done.
remote: Compressing objects: 100% (118/118), done.
remote: Total 143 (delta 7), reused 71 (delta 3), pack-reused 0
Receiving objects: 100% (143/143), 85.57 KiB | 551.00 KiB/s, done.
Resolving deltas: 100% (7/7), done.
Available targets:

[.build-harness]: Clone of build-harness is up-to-date, skipping auto-init
  aws/install                         Install aws cli bundle
```

Nuru: This is me being cautious and conservative. Any exiting scripted docker image knows to call `make init` first. I want to limit this enhancement to being a convenience for interactive use and not have scripts rely on it. I don't know why you would want to run this in a `python:3` container, but if you do, you still have 2 options:

```
$ make init && \
  docker run --rm -ti -v "$PWD:/workdir" -w /workdir --entrypoint make python:3 help
```
or
```
$  docker run --rm -ti -v "$PWD:/workdir" -w /workdir --entrypoint bash python:3 -c 'make init && make help'

Comment (lorengordon) [reply templates/Makefile.build-harness]: Of course `python:3` is contrived. The actual use case I've encountered is when using a CI service that itself runs in a container, e.g. AWS CodeBuild. To force-off the auto-init in that use case, we're now introducing a difference between how you run locally and how you run in CI. That kind of thing is confusing and frustrating in any project. Considering `BUILD_HARNESS_AUTO_INIT=true`, the intention is quite explicit, and of course the user can also disable that.

Anyway, that's the case I'm making. If you insist, I'll go ahead and disable it entirely for docker containers. (Of course, other container runtimes will still auto-init, because they don't generally match the `docker` pattern used in the grep.)

Comment (Nuru) [reply templates/Makefile.build-harness]: Well, I guess you have to pick which "confusing and frustrating" behavior you prefer. Prior to your proposal, everything was consistent with regards to when you did or did not need to run `make init`. Your case for auto-init is that "people kept forgetting to run `make init`", and I am willing to allow your automation for "people" where they can supervise it. (There remains a confusing and frustrating change in behavior if you have the `build-harness` project checked out in a sibling directory, but that is not something we are going to address in this PR.)

I intentionally do not want to change the behavior of CI services because we have a lot of those installed and running and sometimes they do not call `make init` for security reasons. I do not want to create a security vulnerability based on the value of an environment variable. 

What might work and you might find preferable is to add a new pattern rule that invokes init first:

```make
%.init: init
	$(MAKE) $*
```

Then you can run `make readme.init` to auto-init before making readme. (I think it will work, but I can think of a few reasons it would not work, so test it out.)

lorengordon: It's funny, before actually implementing this feature, I was making the same arguments to my team that you are making to me now. Then I started using it and I realized they were right. This feature provides a vastly superior experience. I'm confident that if you use it, you'll come around also.

Dockerfile
lorengordon: Done

templates/Makefile.build-harness
lorengordon: Made the requested change

Nuru (approved):

Nuru:
@lorengordon Just to illustrate why I was so cautious about this, it turns out this PR introduced a kind of serious bug that I missed in my code review. The PR moved `BUILD_HARNESS_PATH :=` to after where it was referenced by `BUILD_HARNESS_DOCKER_SHA_TAG_CMD :=` which caused all the rules that refer to a build-harness Docker image by git SHA to reference the wrong git repo. Broke a lot of stuff. (See _unintended consequences_ above)

lorengordon:
@Nuru Glad you were able to figure it out. Looks like the fix was pretty simple.

PR #273, opened 2021-01-19
feat: add dynamic update the list of related references

SweetOps:
## what

* add script that will dynamically update the list of related references in the `README.yaml`

## references
* CPCO-361


modules/readme/update_related_references.py
```diff
@@ -0,0 +1,92 @@
...
+import json
+import subprocess
+from github import Github
+import requests
+from iteration_utilities import unique_everseen
+from ruamel.yaml import YAML
+
+GH_TOKEN = os.environ["GH_TOKEN"]
+GH_ORG_NAME = os.getenv("GH_ORG_NAME", "cloudposse")
+GH_SEARCH_PATTERN = os.getenv("GH_SEARCH_PATTERN", "terraform-aws-eks")
```
osterman: You'll want to compute this dynamically by looking at the first 3 fields of this module's github repo name. 

e.g. a repo like `https://github.com/cloudposse/terraform-aws-vpc-peering-multi-account`

the first 3 fields are `terraform-aws-vpc` and the pattern would be `terraform-aws-vpc-.*`

osterman: (never mind - looks like you've done it already down below).

In this case, let's set the `GH_SEARCH_PATTERN` to something broad like `.*` (if using regex)

osterman:
```suggestion
GH_SEARCH_PATTERN = os.getenv("GH_SEARCH_PATTERN", "terraform-")
```

bin/generate_related_references.py
```diff
@@ -0,0 +1,94 @@
...
+            name_pattern = "terraform-{}-{}".format(r["provider"], r["name"])
+        item_object["name"] = name_pattern
+        item_object["description"] = r["description"]
+        item_object["url"] = r["source"]
+        items.append(item_object)
+    return items
+
+
+if __name__ == "__main__":
+    related_reference_list = []
```
osterman: So `related` and `references` are two separate sections of the README.

The `related` is for modules relates to this one.

The `references` section is for links to documentation (links to providers, link to AWS docs, etc) to read more, for example.

osterman:
Can you use your branch to open a PR against some of our modules so we can validate the results?

E.g. `terraform-aws-eks`

SweetOps:
> Can you use your branch to open a PR against some of our modules so we can validate the results?

https://github.com/cloudposse/terraform-aws-eks-cluster/pull/98 generated by command:

```sh
docker run --rm -e GH_TOKEN=change_me \
-e GH_SEARCH_PATTERN=terraform-aws-eks \
-v /local_path_to_repo/terraform-aws-eks-cluster:/home/test \
build-harness:test readme/generate-related-reference
```

osterman (approved):
LGTM!

PR #278, opened 2021-02-08
[WIP] Nightly updates for Terraform modules

Nuru:
## what
- Nightly scriptable cron job to update Terraform modules

## why
- Flexibility to add/remove workflows, make other changes automatically

## TODO
- Update null-label modules used by the module (other than `this` in `context.tf`)
- Check for an existing PR
  - Update it if it exists
  - Close it if it no longer has relevant change (e.g. changes already applied manually or via Renovate)


Nuru:
This really isn't necessary anymore. We can hijack `auto-readme` to run whatever else we want to run daily, and in practice we don't generally want to run things daily because of the overhead of dealing with all the resulting PRs.

PR #279, opened 2021-02-10
auto-release respect 'no-release' label

maximmi:
## what
* auto-release respect 'no-release' label

## why
* to be able to avoid publishing a release when one is not needed (ex. when only readme/docs was updated)

### DIFF
--- a/templates/.github/workflows/auto-release.yml
+++ b/templates/.github/workflows/auto-release.yml
@@ -3,17 +3,23 @@ name: auto-release
 on:
   push:
     branches:
-    - master
+      - master
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


osterman (changes requested):
Use the official way: https://github.com/release-drafter/release-drafter#exclude-pull-requests

Nuru:
> Use the official way: https://github.com/release-drafter/release-drafter#exclude-pull-requests

@osterman My understanding is that only excludes the PR message from the release notes. It does not influence whether or not the release is published. I want to completely avoid creating a release when there are only non-code changes, for just one example. I think this PR is the right way to do that.

osterman (approved):

PR #282, opened 2021-02-23
integration of generate-related-references target

maximmi:
## what
* target to generate `related` and `references` sections to README.yaml integrated into auto-format workflow

## why
* to have up to date `related` and `references` sections in README.md on each PR update

### DIFF
--- a/templates/terraform/.github/workflows/auto-format.yml
+++ b/templates/terraform/.github/workflows/auto-format.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   auto-format:
     runs-on: ubuntu-latest
-    container: cloudposse/build-harness:slim-latest
+    container: cloudposse/build-harness:latest
     steps:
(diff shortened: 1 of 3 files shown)


goruha (approved):

templates/terraform/.github/workflows/auto-format.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   auto-format:
     runs-on: ubuntu-latest
-    container: cloudposse/build-harness:slim-latest
+    container: cloudposse/build-harness:latest
```
Nuru, line 9: @maximmi @goruha Since the `auto-format.yml` is installed in every Terraform repo, and we do not currently have an automated way of updating it, this change (or the need for this change) should have alerted you to the fact that committing this PR would break auto-format in every repo.

PR #284, opened 2021-02-24
Add `kubectl@cloudposse` to Dockerfile

aknysh:
## what
* Add `kubectl@cloudposse` to Dockerfile

## why
* Codefresh pipelines use `cloudposse/build-harness` Docker image to deploy helmfile using `kubectl` to switch context
* No need to add `kubectl` to all pipeline steps

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ RUN curl -fsSL --retry 3 https://apk.cloudposse.com/install.sh | bash
 RUN apk --update --no-cache add \
       chamber@cloudposse \
       gomplate@cloudposse \
+      kubectl@cloudposse \
       helm@cloudposse \
       helmfile@cloudposse \
       codefresh@cloudposse \
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2017-2019 Cloud Posse, LLC
+   Copyright 2017-2021 Cloud Posse, LLC
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ In general, PRs are welcome. We follow the typical "fork-and-pull" Git workflow.
 
 ## Copyrights
 
-Copyright © 2016-2021-2021 [Cloud Posse, LLC](https://cloudposse.com)
+Copyright © 2016-2021 [Cloud Posse, LLC](https://cloudposse.com)
 
 
 
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ In general, PRs are welcome. We follow the typical "fork-and-pull" Git workflow.
 
 ## Copyrights
 
-Copyright © 2016-2021 [Cloud Posse, LLC](https://cloudposse.com)
+Copyright © 2016-2021-2021 [Cloud Posse, LLC](https://cloudposse.com)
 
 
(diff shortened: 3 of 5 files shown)


Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN curl -fsSL --retry 3 https://apk.cloudposse.com/install.sh | bash
 RUN apk --update --no-cache add \
       chamber@cloudposse \
       gomplate@cloudposse \
+      kubectl@cloudposse \
```
Nuru, line 39: `kubectl` has to be version matched to the cluster it is controlling. It probably should be added on-demand [like most packages](https://github.com/cloudposse/build-harness/blob/d681457cf521c76a943ebbebb8216f6743b0e887/modules/packages/Makefile#L24). If you are going to pre-install it, you need to install all the versions and then select the right one [the way we do in Geodesic](https://github.com/cloudposse/geodesic/blob/dfba4fd41620371ce091423a7bc7e8a1ce2154fc/rootfs/usr/local/bin/kubectl-auto-select)

```suggestion
      kubectl@cloudposse-1.16 \
      kubectl@cloudposse-1.17 \
      kubectl@cloudposse-1.18 \
      kubectl@cloudposse-1.19 \
```

Nuru (changes requested):
`kubectl` has to be version matched to the cluster it is controlling. It probably should be added on-demand [like most packages](https://github.com/cloudposse/build-harness/blob/d681457cf521c76a943ebbebb8216f6743b0e887/modules/packages/Makefile#L24). If you are going to pre-install it, you need to install all the versions and then select the right one [the way we do in Geodesic](https://github.com/cloudposse/geodesic/blob/dfba4fd41620371ce091423a7bc7e8a1ce2154fc/rootfs/usr/local/bin/kubectl-auto-select)

Dockerfile
aknysh: we just need a simple `kubectl` command to run in the pipelines: `kubectl config use-context ${{KUBE_CONTEXT}}`
All of what we do in `geodesic` is too complicated.
Also, we need `kubectl` to run `config use-context ` to be able to connect to the cluster, so we can't determine the server version before calling `kubectl` (this remains an issue regardless where we install it, in `build-harness` or by adding it in the pipeline steps)

osterman: Geodesic just a toolbox. Why can't we use it?

osterman: That way the customer uses the same versions of the tools. They are not dependent on us. We can't break their release engineering process.

aknysh: ok, we can use `geodesic` (we just discussed before to use `build-harness`, now I'm not sure why)

Nuru: I'm tired and cranky, so I will say
1. `helm` doesn't need `kubectl`, so we probably either don't need `kubectl` or should have the right version
2. The call to the Kubernetes API to determine the version is not subject to change the way the rest of the API is. (Ongoing changes to the Ingress resource highlight why it is important to have the right version of `kubectl` if it is going to have anything to do with them.) Basically you can just `curl ${SERVER}/version` where `SERVER` is the `server` in the selected context in the KUBECONFIG file. 
3. On the other hand, having the wrong version of `kubectl` installed leads to subtle and hard-to-find errors when it does break. I would prefer to head that off by installing the right version to begin with.

We can update the Geodesic script to use `yq` and `jq` to parse the Kubeconfig file and fetch the version via `curl`, then we can use the package installer to install the right version.

Nuru: Agreed, much better to use Geodesic for CD deploy

PR #288, opened 2021-04-17
Add support for "deprecated" flag/section in README

Nuru:
## what
- Add support for "deprecated" flag/section in README.yaml -> README.md
- Update `auto-release` workflow to run on push to other default branch names besides `master`

## why
- Make it easy and consistent to mark a module as deprecated
- Allow us to use the same YAML file in all our repos as we change our default branch name

## example

Add this to `README.yaml`

```yaml
deprecated: 
  notice: |-
    This module is obsolete and has been replaced with [something else](link to something else)

```

or simply
```yaml
deprecated: true
```

### DIFF
--- a/templates/.github/workflows/auto-release.yml
+++ b/templates/.github/workflows/auto-release.yml
@@ -3,7 +3,9 @@ name: auto-release
 on:
   push:
     branches:
+      - main
       - master
+      - production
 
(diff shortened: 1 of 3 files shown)


Gowiem (approved):
👍

PR #292, opened 2021-06-11
Add possible disable Security & Compliance section

ByJacob:
## what
* Add possible disable Security & Compliance section

## why
* Somebody want use this tool and don't required Security & Compliance secion

## references
* Don't create issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,13 +171,13 @@ Available targets:
   geodesic/deploy                     Run a Jenkins Job to Deploy $(APP) with $(CANONICAL_TAG)
   git/aliases-update                  Update git aliases
   git/export                          Export git vars
-  git/submodules-update               Update submodules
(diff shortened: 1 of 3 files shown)


ByJacob:
@Nuru can we process this MR ?

Nuru:
@ByJacob Sorry about the delay in getting to this. 

We have about 100 modules that rebuild their README's every night, and when the README's change, that triggers a new release. Eventually we will clean that up so that (a) changes in whitespace only do not trigger updates and (b) updates to the README do not trigger new releases, but until then, any changes to the `templates/README.md.gotmpl` must produce *exactly* the same output or else we get 100 releases of whitespace changes, which we really do not want.

Also, disabling our branding is not something we are keen on spending time on, so I am going to close this as "wontfix". Feel free to use your own fork for this if you want to.

PR #296, opened 2021-08-24
Create a draft release rather than no release with `no-release` label

Nuru:
## what
- When a PR is labeled `no-release`, still create a draft release

## why
- Allows you to accumulate releases via the `no-release` tag and then make a manual release out of the draft release, taking advantage of the auto-releaser's drafting capabilities.


aknysh (approved):

danjbh (approved):
LGTM

PR #299, opened 2021-11-11
Make 1.x the default Terraform version

Nuru:
## what
- Make 1.x the default Terraform version, replacing 0.13
- Add  a `build` target to the Makefile

## why
- When running the `build-harness` Docker image manually, tools usually use the default Terraform version, and this affects things like Terraform formatting. Cloud Posse is now using Terraform 1.x on all new projects, so it is a better default.
- The plain `docker/build` `make` target does not tag the Docker image with the tag other `make` targets expect to find, which makes developing `build-harness` unnecessarily difficult. Now `make build` will build and tag the image so that other commands like `make build-harness/shell` will find the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN apk --update --no-cache add \
     sed -i /PATH=/d /etc/profile
 
 # Use Terraform 0.13 by default
-ARG DEFAULT_TERRAFORM_VERSION=0.13
+ARG DEFAULT_TERRAFORM_VERSION=1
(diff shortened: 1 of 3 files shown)


nitrocode (approved):

Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN apk --update --no-cache add \
     sed -i /PATH=/d /etc/profile
 
 # Use Terraform 0.13 by default
```
aknysh, line 59:
```suggestion
# Use Terraform 1.x.x by default
```

aknysh (commented):
a nitpick

PR #301, opened 2021-11-17
switch to using terraform-1.0 instead of terraform-1

mcalhoun:
## what
-  Switch to using the terraform-1.0 package rather than the terraform-1 package

## why

- Some scripts expect Terraform to be installed in a directory named with major.minor and the previous terraform-1 installed to major only.

## references

- This should not be merged until after cloudposse/packages#1848 is merged

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,32 +7,32 @@ LABEL "com.github.actions.icon"="tool"
 LABEL "com.github.actions.color"="blue"
 
 RUN apk --update --no-cache add \
-      bash \
-      ca-certificates \
-      coreutils \
-      curl \
-      git \
-      gettext \
-      go \
-      grep \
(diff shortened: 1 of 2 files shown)


Nuru:
Creating and using a `terraform-1.0` package was explicitly rejected in https://github.com/cloudposse/packages/pull/1848 because in the SemVer standard `0.x` is not `major`.`minor` it is `0.major`. Scripts that have a problem with this should be fixed to work with the correct `major` version number.

Nuru:
@mcalhoun Hijacking this PR to provide backward support for the Makefiles, which was your goal, via symlinks, which was not your method.

Nuru:
Oops, can't hijack it because I force pushed while it was closed. Who knew? Bummer.

PR #302, opened 2021-11-18
 Finish #299, default to Terraform v1 in builder-slim

Nuru:
## what

- Finish #299, default to Terraform v1 in builder-slim
- Symlink current Terraform `major`.`minor` directories to `major`
- Fix bash test. Thanks @tevansuk
- Support `no-release` label

## why

* #299 Made Terraform 1.x for the `builder` Docker image but neglected to update the `builder-slim` Docker image.
* Some deployed Makefiles expect Terraform to be installed in a directory named with `major`.`minor` due to pre-production versions using semver `0.major` versioning. With Terraform production version, the major version is back to its rightful place, so we only need `major`, but the old Makefile are installed in hundreds of Terraform modules, so we need some backward support for them.
* Bad syntax in Makefile causes errors during auto-init.
* Sometimes we want to merge changes (such as for PR #298) without triggering a release, for example to bundle several PRs into a single release.

## references

* Incorporates and closes #298. Thank you @tevansuk

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -9,10 +9,15 @@ jobs:
   semver:
     runs-on: ubuntu-latest
     steps:
+    # Get PR from merged commit to master
+    - uses: actions-ecosystem/action-get-merged-pull-request@v1
+      id: get-merged-pull-request
+      with:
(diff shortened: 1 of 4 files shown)


mcalhoun (approved):
LGTM!

PR #310, opened 2022-02-18
GitHub auto update

Nuru:
## what
- Update GitHub workflows as part of make pr/auto-format
- Update `codeowners-validator` to current version 0.7.1
- Update Go 1.15.11 -> 1.17.7, Alpine 3.13 -> 3.15

## why
- Keep modules up-to-date without triggering massive waves of PRs that require high-level approvals
- Version 0.5.0 stopped working
- Security fixes

### DIFF
--- a/modules/github/Makefile.init
+++ b/modules/github/Makefile.init
@@ -17,16 +17,21 @@ GITHUB_TERRAFORM_TEMPLATES = .github/workflows/chatops.yml \
 	.github/mergify.yml \
 	.github/renovate.json
 
+# Create a file (contents unimportant) with this name
+# to prevent pr/auto-format etc. from updating the GitHub templates.
+GITHUB_UPDATE_DISABLE_SENTINEL := .github/.github-update-disabled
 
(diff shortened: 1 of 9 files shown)


aknysh (commented):
update LICENSE as well https://github.com/cloudposse/build-harness/blob/master/LICENSE#L189

aknysh (approved):

modules/github/Makefile.init
```diff
@@ -17,16 +17,21 @@ GITHUB_TERRAFORM_TEMPLATES = .github/workflows/chatops.yml \
 	.github/mergify.yml \
 	.github/renovate.json
 
+# Create a file (contents unimportant) with this name
+# to prevent pr/auto-format etc. from updating the GitHub templates.
+GITHUB_UPDATE_DISABLE_SENTINEL := .github/.github-update-disabled
```
osterman, line 22: What I don't like about this is that it's not specific to any workflow. It's like a ENV that we have no idea what consumes it. What about naming it after the workflow it disables?

```suggestion
GITHUB_UPDATE_DISABLE_SENTINEL := .github/.github-update-disabled
```

Nuru: It doesn't disable any workflow. It disables the `make github/update` target.

PR #312, opened 2022-03-10
Auto-readme workflow fix: name of default branch rather than hard coding it

Nuru:
## what
- Fix `auto-readme` workflow to detect the name of the default branch rather than rely on hard coded `main`
- Only output the message "Ignoring whitespace-only changes in README" when, in fact, there are changes to ignore

## why
- Most of our repos still use `master` for the default branch, but some use `main` and we need the right target for the PR this workflow creates
- It was confusing to see that message when there were, in fact, no changes at all

### DIFF
--- a/templates/terraform/.github/workflows/auto-readme.yml
+++ b/templates/terraform/.github/workflows/auto-readme.yml
@@ -1,5 +1,7 @@
 name: "auto-readme"
 on:
+  workflow_dispatch:
+
   schedule:
   # Example of job definition:
   # .---------------- minute (0 - 59)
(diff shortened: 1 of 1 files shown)


templates/terraform/.github/workflows/auto-readme.yml
```diff
@@ -15,21 +17,35 @@ on:
...
 
+    - name: Find default branch name
+      id: defaultBranch
+      shell: bash
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      run: |
+        default_branch=$(gh repo view --json defaultBranchRef --jq .defaultBranchRef.name)
+        printf "::set-output name=defaultBranch::%s\n" "${default_branch}"
+        printf "defaultBranchRef.name=%s\n" "${default_branch}"
```
osterman, line 33: Why are we just not using https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/

There's native support for this in GHA.

Nuru: `$default-branch` is a macro expanded when copying a template "starter workflow" to a destination repository. It does not work with our make-based file copies. From [the documentation](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization#creating-a-starter-workflow):
> If you need to refer to a repository's default branch, you can use the $default-branch placeholder. _When a workflow is created the placeholder will be automatically replaced with the name of the repository's default branch_. (Emphasis added.)

It does not work inside a deployed workflow.

Nuru: We could possibly use `${{ github.event.repository.default_branch }}` but the way I read the [documentation of the "schedule" event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) is that `github.event` is not available when a scheduled run takes place.

osterman (approved):

templates/terraform/.github/workflows/auto-readme.yml
osterman: Ok, let's stick with what you have....

Posting this for reference https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable

Probably should just create our own action for this. I kind of like the API approach in the last example ☝️

Issue #314, opened 2022-04-26
DEPRECATION NOTICE: git.io is shutting down (update your links!)


osterman:
#  GIT.IO DEPRECATION

## What 

Remove all references to `git.io/build-harness` and refer to`cloudposse.tools/build-harness` instead.

See [our Git.io deprecation documentation](https://github.com/cloudposse/build-harness/blob/master/docs/git-io-deprecation.md) for more details, including instructions on how to update.

## Why

GitHub announced that the [`git.io` redirector service is shutting down on 2022-04-29](https://github.blog/changelog/2022-04-25-git-io-deprecation/). GitHub only provided 4 days notice of this shutdown.

This means all references to `git.io/build-harness` will stop working.

## Update

GitHub amended their decision to completely shut down `git.io` (and amended the above linked [blog post](https://github.blog/changelog/2022-04-25-git-io-deprecation/))), but you should still make the change when convenient in order to preserve functionality going forward.

Nuru:
Fixed in #315 and #317, updated in #319

PR #315, opened 2022-04-26
Use cloudposse.tools instead of git.io

osterman:
## what
* Update to use `cloudposse.tools` instead of `git.io`

## why
* `git.io` is shutting down 2022-04-29

## references
* #314

### DIFF
--- a/README.yaml
+++ b/README.yaml
@@ -1,5 +1,5 @@
 ---
-#
+					#
 # This is the canonical configuration for the `README.md`
 # Run `make readme` to rebuild the `README.md`
 #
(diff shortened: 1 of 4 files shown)


osterman:
Waiting on DNSSEC propagation for cloudposse.tools

Benbentwo (dismissed):

nitrocode (dismissed):

README.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-#
+					#
```
Nuru:
```suggestion
#
```

Nuru (changes requested):

Nuru (approved):

PR #318, opened 2022-04-26
Remove duplicate paragraph

Nuru:
## what
* Remove duplicate paragraph in README from #316 
* Update `find` command to exclude symbolic links


osterman (approved):
Would prefer we move the long-form instructions to another file

PR #327, opened 2022-08-15
Added github action docs

goruha:
## what
* Added Github action docs

## why
* Generated README docs from `action.yaml`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,13 @@ RUN apk --update --no-cache add \
       boto==2.49.0 \
       boto3==1.21.1 \
       iteration-utilities==0.11.0 \
-      pre-commit \
       PyGithub==1.55 && \
     git config --global advice.detachedHead false
+
+## Workaround https://github.com/pypa/pip/issues/5247
+RUN pip3 install --upgrade --force-reinstall pip==9.0.3 && \
+    pip3 install --no-cache-dir --disable-pip-version-check pre-commit && \
+    pip3 install --upgrade pip
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN curl -fsSL --retry 3 https://apk.cloudposse.com/install.sh | bash
--- a/docs/git-io-deprecation.md
+++ b/docs/git-io-deprecation.md
@@ -63,7 +63,7 @@ Once you are happy with the command to generate the list of files to update, upd
 sed -i '' 's/git.io\/build-harness/cloudposse.tools\/build-harness/' $(<command to list files>)
 ```
 
-#### Examples
+#### Samples
 
 The quickest update will be if you only have a single project to update, in which case you can `cd` into the project root directory and
 ```bash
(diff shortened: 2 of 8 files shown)


aknysh (approved):

Dockerfile
```diff
@@ -31,10 +31,14 @@ RUN apk --update --no-cache add \
...
       iteration-utilities==0.11.0 \
-      pre-commit \
       PyGithub==1.55 && \
     git config --global advice.detachedHead false
 
+## Workaround https://github.com/pypa/pip/issues/5247
+RUN pip3 install --upgrade --force-reinstall pip==9.0.3 && \
+    pip3 install --no-cache-dir --disable-pip-version-check pre-commit && \
+    pip3 install --upgrade pip
+
```
Nuru, line 41: Why is this needed? The referenced issue is for Pip 9 and 10 and Python 2.7. We are using Pip v22 and Python 3.9. Any problem this is fixing should probably be fixed some other way.

docs/git-io-deprecation.md
```diff
@@ -39,44 +39,44 @@ or a `src` (or similar) directory under which you have all your `git` repos (dir
...
 ### Updating the affected files
 
 Once you are happy with the command to generate the list of files to update, update the files by inserting that command into this command template:
-```
+```bash
 sed -i '' 's/git.io\/build-harness/cloudposse.tools\/build-harness/' $(<command to list files>)
 ```
 
-#### Examples
+#### Samples
```
Nuru, line 66: Where did this change come from? I don't know what was wrong with "Examples" but I really don't like "Samples". How about "Options"?
```suggestion
#### Options
```

Nuru (changes requested):
I don't know how or why "examples" got changed to "samples", but I don't like it. Can we please either revert it or change it to "options"?

I would like to see the output of the new github actions documentation. Let's please run it and also make sure it gets added to the "make readme" for the project.

Dockerfile
goruha: But the problem was still there 
https://github.com/cloudposse/build-harness/runs/7847715248?check_suite_focus=true

```
docutils, colorama, cfgv, virtualenv, pynacl, deprecated, botocore, s3transfer, PyGithub, pre-commit, boto3, awscli
#5 32.27   Attempting uninstall: distlib
#5 32.27     Found existing installation: distlib 0.3.3
#5 32.27 ERROR: Cannot uninstall 'distlib'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

goruha: @Nuru ^

docs/git-io-deprecation.md
goruha: https://github.com/cloudposse/build-harness/runs/7840582188?check_suite_focus=true

```
/github/workspace/README.md:390 MD024/no-duplicate-heading/no-duplicate-header Multiple headings with the same content [Context: "#### Examples"]
```

I think that `Samples` means the same as  `Examples`, does not it?

goruha: @Nuru ^

PR #331, opened 2022-10-03
update github action template

mcalhoun:
## what
* Update the Github Action docs template

## why
* So the template doesn't error if the `inputs` or `outputs` don't exist for an action


danjbh (approved):
lgtm!

PR #332, opened 2022-10-19
Replace `set-output` in Templates

milldr:
## what
- Replaced set-output in terraform workflow templates

## why
- The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.

## references
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


zdmytriv (approved):
thanks for fixing it

PR #335, opened 2023-01-12
fix(modules/git/bootstrap): remove lazy assignments

### LINKED ISSUE #333: make 4.4 (oct 31 2022) causes issues with the git submodule resulting in an infinite loop
Found a bug? Maybe our [Slack Community](https://slack.cloudposse.com) can help. 

[Slack Community](https://slack.cloudposse.com)

## Describe the Bug
make 4.4 (oct 31 2022) and 4.4.1 causes issues with the git submodule resulting in an infinite loop

https://ftp.gnu.org/gnu/make/

cc: @Nuru 

## Expected Behavior
It should not be in an infinite loop

## Steps to Reproduce

1. Use the ancient OSX make which is 3.81 version
1. `/usr/bin/make init`
1. Notice that the target exits successfully
1. Use the GNU make by doing `brew install make` to get the latest 4.4 version
1. `/usr/local/bin/make init`
1. Notice the infinite loop

```sh
✗ make SHELL='sh -x' init

+ pwd
+ pwd
+ '[' -d build-harness ']'
+ pwd
+ config --get remote.origin.url
+ sed 's/\.git$//g'
+ sed 's/git@\(.*\):/https:\/\/\1\//g'
+ show -s --format=%aN
+ show -s --format=%aN
+ config --get remote.origin.url
+ sed 's/\.git$//g'
+ sed 's/git@\(.*\):/https:\/\/\1\//g'
+ rev-parse --verify --short HEAD
+ config --get remote.origin.url
+ sed 's/\.git$//g'
+ sed 's/git@\(.*\):/https:\/\/\1\//g'
+ rev-parse --verify --short HEAD
+ rev-parse --verify --short HEAD
+ config --get remote.origin.url
```

## Screenshots
N/A

## Environment (please complete the following information):
- OSX Ventura 13.0
- GNU Make 4.4

## Additional Context
Worked on debugging this with @dudymas on it! Thank you!

Downgrading back to make 4.3 is successful

Commenting out this line prevents the issue

https://github.com/cloudposse/build-harness/blob/72ee8e2eb3525224ab5d6093d4f919d572eef8cd/Makefile#L42

And the `git` module is the specific one that seems to be failing which causes repeating the include.

https://github.com/cloudposse/build-harness/blob/72ee8e2eb3525224ab5d6093d4f919d572eef8cd/modules/git/bootstrap.Makefile#L11-L17

### Option 1: Use OSX make

```shell
/usr/bin/make --version
/usr/bin/make all
```

### Option 2: Brew to pin GNU Make to 4.3

if you do not want to use the older make but also want to pin to 4.3 using homebrew

- [Brew make.rb raw 4.3 formula](https://raw.githubusercontent.com/Homebrew/homebrew-core/9345e061435f18a91437cc5a3db34a90acbb9f1b/Formula/make.rb)
- [Brew make.rb formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/make.rb)

```sh
# uninstall 4.4 make
brew uninstall make
# find your make.rb file
find / -name make.rb
# override the make.rb file with the 4.3 version
wget https://raw.githubusercontent.com/Homebrew/homebrew-core/9345e061435f18a91437cc5a3db34a90acbb9f1b/Formula/make.rb make.rb
# install
brew install make.rb
# finally pin
brew pin make
```

dudymas:
## what
* remove recursive assignments for git variables

## why
* recursive assignments cause race conditions
* none of the git bootstrap operations are costly to calculate
* make 4.4 exhibits a complete deadlock due to all the recursive assignments

## references
* this fixes #333 
* https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html

### DIFF
--- a/modules/git/bootstrap.Makefile
+++ b/modules/git/bootstrap.Makefile
@@ -4,28 +4,28 @@ ifeq ($(wildcard .git),)
   endif
 else
 
-GIT ?= $(shell which git)
+GIT := $(shell which git)
 
-export GIT_COMMIT ?= $(shell $(GIT) rev-parse --verify HEAD 2>/dev/null)
-export GIT_COMMIT_SHORT ?= $(shell $(GIT) rev-parse --verify --short HEAD 2>/dev/null)
-export GIT_BRANCH ?= $(shell $(GIT) rev-parse --abbrev-ref HEAD 2>/dev/null)
-export GIT_TAG ?= $(shell $(GIT) tag -l --sort -taggerdate --points-at HEAD 2>/dev/null | head -n 1)
-export GIT_COMMIT_URL ?= $(shell $(GIT) config --get remote.origin.url 2>/dev/null | sed 's/\.git$$//g' | sed 's/git@\(.*\):/https:\/\/\1\//g' )/commit/$(GIT_COMMIT_SHORT)
+export GIT_COMMIT := $(shell $(GIT) rev-parse --verify HEAD 2>/dev/null)
+export GIT_COMMIT_SHORT := $(shell $(GIT) rev-parse --verify --short HEAD 2>/dev/null)
+export GIT_BRANCH := $(shell $(GIT) rev-parse --abbrev-ref HEAD 2>/dev/null)
+export GIT_TAG := $(shell $(GIT) tag -l --sort -taggerdate --points-at HEAD 2>/dev/null | head -n 1)
+export GIT_COMMIT_URL := $(shell $(GIT) config --get remote.origin.url 2>/dev/null | sed 's/\.git$$//g' | sed 's/git@\(.*\):/https:\/\/\1\//g' )/commit/$(GIT_COMMIT_SHORT)
 
-export GIT_COMMIT_MESSAGE ?= $(shell $(GIT) show -s --format=%s%b 2>/dev/null)
-export GIT_COMMIT_AUTHOR ?= $(shell $(GIT) show -s --format=%aN 2>/dev/null)
-export GIT_COMMIT_TIMESTAMP ?= $(shell $(GIT) log -1 --format=%ct 2>/dev/null)
+export GIT_COMMIT_MESSAGE := $(shell $(GIT) show -s --format=%s%b 2>/dev/null)
+export GIT_COMMIT_AUTHOR := $(shell $(GIT) show -s --format=%aN 2>/dev/null)
+export GIT_COMMIT_TIMESTAMP := $(shell $(GIT) log -1 --format=%ct 2>/dev/null)
 
 
 ## GIT_TIMESTAMP is deprecated. Use GIT_COMMIT_TIMESTAMP instead
-export GIT_TIMESTAMP ?= $(shell $(GIT) log -1 --format=%ct 2>/dev/null)
+export GIT_TIMESTAMP := $(shell $(GIT) log -1 --format=%ct 2>/dev/null)
 
 ifeq ($(GIT_TAG),)
   export GIT_IS_TAG := 0
-  export GIT_BRANCH_TAG ?= $(GIT_BRANCH)
+  export GIT_BRANCH_TAG := $(GIT_BRANCH)
 else
   export GIT_IS_TAG := 1
-  export GIT_BRANCH_TAG ?= $(GIT_TAG)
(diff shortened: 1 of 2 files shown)


modules/git/bootstrap.Makefile
```diff
@@ -34,8 +34,8 @@ else
   export GIT_IS_BRANCH := 1
 endif
 
-export GIT_URL ?= $(shell git remote get-url origin 2>/dev/null)
```
dudymas, line 37: I noticed this uses raw `git` instead of `$(GIT)` ... I'm leaving it as is, but if folks think that's a mistake, I can change it. @mcalhoun last touched this code, AFAIK... so maybe they have ideas?

nitrocode: It seems like a mistake to not use `$(GIT)`

modules/git/bootstrap.Makefile
```diff
@@ -4,28 +4,28 @@ ifeq ($(wildcard .git),)
   endif
 else
 
-GIT ?= $(shell which git)
+GIT := $(shell which git)
```
nitrocode, line 7: Is there no way to fix this for make 4.4 and retain the `?=` form?

modules/git/bootstrap.Makefile
```diff
@@ -4,28 +4,28 @@ ifeq ($(wildcard .git),)
...
 
-export GIT_COMMIT ?= $(shell $(GIT) rev-parse --verify HEAD 2>/dev/null)
```
nitrocode, line 9: How come only this module is affected by the make 4.4 issue and not other modules that also use the `?=` format?

modules/git/bootstrap.Makefile
```diff
@@ -4,28 +4,28 @@ ifeq ($(wildcard .git),)
...
-export GIT_COMMIT_SHORT ?= $(shell $(GIT) rev-parse --verify --short HEAD 2>/dev/null)
-export GIT_BRANCH ?= $(shell $(GIT) rev-parse --abbrev-ref HEAD 2>/dev/null)
-export GIT_TAG ?= $(shell $(GIT) tag -l --sort -taggerdate --points-at HEAD 2>/dev/null | head -n 1)
-export GIT_COMMIT_URL ?= $(shell $(GIT) config --get remote.origin.url 2>/dev/null | sed 's/\.git$$//g' | sed 's/git@\(.*\):/https:\/\/\1\//g' )/commit/$(GIT_COMMIT_SHORT)
+export GIT_COMMIT := $(shell $(GIT) rev-parse --verify HEAD 2>/dev/null)
```
nitrocode, line 9: Is it possible that the `?=` form is triggering a bug in make 4.4? If so, would you create an upstream issue with a minimum viable reproducible example so it can be fixed in make 4.5?

modules/git/bootstrap.Makefile
dudymas: I can try doing something like:
```
PATHWISE_GIT := $(shell which git)
GIT ?= $(PATHWISE_GIT)
```

but is that realistically something we'd want to give folks configuration over? that is to say, I'd question when would folks need to override the path of git? And if they did so, would there be a need to consistently do so?

modules/git/bootstrap.Makefile
osterman: Yea, let's fix this, then merge. We can always do follow up PRs. But this affects too many to ignore.

modules/git/bootstrap.Makefile
dudymas: it could be that it's a behavior in how `bootstrap` level Makefiles get imported.

modules/git/bootstrap.Makefile
dudymas: I believe make 4.4 is the latest stable. It wouldn't necessarily be a good idea to run bleeding edge make, but we can try it. It would have to be built from scratch on each platform we support.

nitrocode (approved):

modules/git/bootstrap.Makefile
nitrocode: I wasn't suggesting that we run a bleeding edge make...

I was suggesting if this is a bug, we should submit it upstream.

PR #340, opened 2023-02-16
Updated CODEOWNERS to own gha in templates

zdmytriv:
## what
* Updated CODEOWNERS so GitHub Actions in templates are also owned by `@cloudposse/engineering` team

## why
* More control by the team on what gets changed

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,11 +4,15 @@
 # Order is important: the last matching pattern takes the most precedence
 
 # These owners will be the default owners for everything
-*             @cloudposse/engineering @cloudposse/contributors
+*                               @cloudposse/engineering @cloudposse/contributors
 
 # Cloud Posse must review any changes to Makefiles
-**/Makefile   @cloudposse/engineering
-**/Makefile.* @cloudposse/engineering
+**/Makefile                     @cloudposse/engineering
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -4,11 +4,15 @@
...
+**/Makefile                     @cloudposse/engineering
+**/Makefile.*                   @cloudposse/engineering
 
 # Cloud Posse must review any changes to GitHub actions
-.github/*     @cloudposse/engineering
+.github/*                       @cloudposse/engineering
+
+# Cloud Posse must review any changes to GitHub templates actions
+templates/.github/*             @cloudposse/engineering
+templates/terraform/.github/*   @cloudposse/engineering
```
zdmytriv, line 18: Alternative might be
```
**/.github/*        @cloudposse/engineering
```

osterman: I like being explicit the way you are doing it now.

osterman (dismissed):

osterman (approved):

.github/CODEOWNERS
Nuru: No, default is local directory, so should be `**/.gomplate.*` (or `.yaml` and `.yml`). Also should be owned by `cloudposse/admin`, no engineering, because it is tricky and we are not using it.

PR #341, opened 2023-02-21
Gomplate to use predefined config

max-lobur:
## what

- Define gomplate config and pass it throught CLI flag to every gomplate invocation
- Passing as the very last option to avoid injecting override in the env vars dereferenced later in same command
- unset GOMPLATE_CONFIG is not required as CLI option always takes precedence over env var, I tested.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,5 +16,5 @@
 # Cloud Posse must review any changes to GitHub templates actions
 templates/.github/*             @cloudposse/engineering
 templates/terraform/.github/*   @cloudposse/engineering
-**/.gomplate.* @cloudposse/admins
-**/configs/gomplate.yaml @cloudposse/admins
+**/.gomplate.*                  @cloudposse/admins
+**/configs/gomplate.yaml        @cloudposse/admins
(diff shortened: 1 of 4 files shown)


max-lobur:
Testing in few repos ATM

modules/readme/Makefile
```diff
@@ -29,8 +29,7 @@ readme/lint:
 
 ## Create README.md by building it from README.yaml
 readme/build: readme/deps $(README_DEPS)
-	@gomplate --file $(README_TEMPLATE_FILE) \
-		--out $(README_FILE) 
+	@gomplate --file $(README_TEMPLATE_FILE) --out $(README_FILE) --config $(BUILD_HARNESS_PATH)/configs/gomplate.yaml
```
osterman, line 32: Does `--config` override `GOMPLATE_CONFIG`?

.github/CODEOWNERS
```diff
@@ -16,3 +16,4 @@
 # Cloud Posse must review any changes to GitHub templates actions
 templates/.github/*             @cloudposse/engineering
 templates/terraform/.github/*   @cloudposse/engineering
+**/.gomplate.*
```
osterman:
```suggestion
**/.gomplate.* @cloudposse/admins
**/configs/gomplate.yaml @cloudposse/admins
```

modules/readme/Makefile
max-lobur: yes, tested this

osterman (approved):

PR #342, opened 2023-03-28
Update templates to use the new shared workflows

max-lobur:
## what

- Update templates to use the new shared workflows
- Add the new make target `precommit/terraform` that as much as possible mimics the new workflow but on the local host 

ATTENTION:
auto-release.yml (config, not workflow) is now terraform specific and moved into terraform subdir. For anything else that uses auto-release build-harness will need to have own corresponding template directory, bc template file override is not supported (make lint fail)

## why

In the scope of the new workflows rollout

## references

https://github.com/cloudposse/github-actions-workflows-terraform-module/tree/main/.github/workflows


max-lobur:
Ready for review

max-lobur:
WIP as I might need to add settings.yaml here first

osterman (dismissed):

modules/terraform/Makefile
```diff
@@ -114,3 +119,6 @@ terraform/rewrite-required-providers: terraform/loosen-constraints
 terraform/v14-rewrite: TERRAFORM = terraform-0.13
 terraform/v14-rewrite: terraform/loosen-constraints terraform/bump-tf-12-min-version terraform/rewrite-required-providers terraform/rewrite-module-source terraform/rewrite-readme-source
 	@{ [[ "$(TERRAFORM_FORCE_README)" != "true" ]] && git diff --no-patch --exit-code README.yaml; } || $(MAKE) readme
+
+
+pr/terraform: terraform/fmt terraform/tflint readme/build
```
osterman: IMO, this is more about `precommit` that PR.

osterman: e.g. `precommit/terraform`

osterman: I could imagine us having a precommit hook that calls this target.

zdmytriv (approved):

PR #345, opened 2023-04-06
chore: deprecate glide in go targets

dudymas:
## what

* deprecate glide
* replace glide targets with go get/mod

## why

* supporting both go mod and glide is tricky


modules/go/Makefile.build
```diff
@@ -18,13 +18,21 @@ else
...
+	$(call assert-set,GO)
+	$(GO) get
+	$(GO) mod tidy
+
+## DEPRECATED Install dependencies
+go/glide-deps: $(GLIDE)
+	@echo "DEPRECATED: Please update to 'go mod' and use go/deps"
 	$(call assert-set,GLIDE)
 	$(GLIDE) install --strip-vendor
 	$(GLIDE) update  --strip-vendor
```
osterman:
```suggestion
	exit 1
```

osterman: Maybe move to `Makefile.deprecated`

modules/go/Makefile.build
```diff
@@ -40,9 +48,14 @@ go/deps-dev: $(GO)
...
+	
+## Clean compiled binary and dependencies
 go/clean-all: go/clean
+	$(GO) mod download
+
+## DEPRECATED Clean compiled binary and dependency
+go/glide-clean-all: go/clean
+	@echo "DEPRECATED: Please update to 'go mod' and use go/clean-all"
 	rm -rf vendor
 	rm -rf glide.lock
```
osterman:
```suggestion
	@exit 1
```

modules/go/Makefile.build
dudymas: Let's just remove it altogether in that case

modules/go/Makefile.build
```diff
@@ -18,18 +18,10 @@ else
...
-	$(GLIDE) update  --strip-vendor
-
-## Install dependencies for build
-go/deps-build:
-	$(call assert-set,GOPATH)
-	mkdir -p  $(GOPATH)/bin
-	# 2022-098-02 https://glide.sh/get no longer works (parked domain, bad SSL cert)
-	# which $(GLIDE) || (curl https://glide.sh/get | sh)
-	which $(GLIDE) || apk add --no-cache glide
+go/deps: $(GO)
```
zdmytriv, line 20: I see `GLIDE` at the very top of this file

aknysh (approved):

PR #348, opened 2023-05-19
Add precommit/terraform make target

Nuru:
## what

- Add `precommit/terraform` `make` target
- Update old `pr/auto-format` `make` target to `make` new `terraform/precommit` target
- Fix `terraform/precommit` target to only run once (not twice)
- Add `tflint` to `build-harness-slim`
- Move `.tflint.hcl` from `/` to `/root` (a.k.a. `$HOME`)
- Update `go` to 1.20.4
- Update Alpine to 3.17

## why

- PR #342 claimed to add `precommit/terraform` target but did not
- Enable people and workflows to run formatting, `tflint`, and other processes locally before opening a PR
- Allow both full and slim Docker images to run `precommit/terraform`
- Move `.tflint.hcl` to where `tflint` expect it to be, in order to support `tflint --enable-plugin=aws`
- Keep frameworks up-to-date

### DIFF
--- a/templates/Makefile.build-harness
+++ b/templates/Makefile.build-harness
@@ -122,12 +122,13 @@ build-harness/shell builder pr/pre-commit pr/readme pr/github-update: RUNNER_DOC
 build-harness/shell builder: build-harness/runner
 	@exit 0
 
-.PHONY: build-harness/shell-slim builder-slim pr/auto-format pr/auto-format/host pr/readme pr/readme/host pr/pre-commit pr/github-update pr/github-update/host tf14-upgrade
+.PHONY: build-harness/shell-slim builder-slim pr/readme pr/readme/host pr/pre-commit pr/github-update pr/github-update/host tf14-upgrade
+.PHONY: precommit/terraform pr/auto-format precommit/terraform/host pr/auto-format/host
 
-build-harness/shell-slim builder-slim pr/auto-format pr/readme tf14-upgrade: RUNNER_DOCKER_IMAGE ?= $(BUILD_HARNESS_DOCKER_IMAGE)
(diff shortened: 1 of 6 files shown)


templates/Makefile.build-harness
```diff
@@ -136,13 +137,19 @@ build-harness/shell-slim builder-slim: ARGS := $(if $(TARGETS),$(TARGETS),-l ||
 build-harness/shell-slim builder-slim: ENTRYPOINT := $(if $(TARGETS),/usr/bin/make,/bin/bash)
 build-harness/shell-slim builder-slim: build-harness/runner
 
-pr/auto-format pr/readme pr/pre-commit pr/github-update tf14-upgrade : ENTRYPOINT := /usr/bin/make
+precommit/terraform pr/auto-format pr/readme pr/pre-commit pr/github-update tf14-upgrade : ENTRYPOINT := /usr/bin/make
```
max-lobur, line 140: The target from the original PR was renamed to `terraform/precommit`. You can call it here , or deprecate . I also noticed that you don't actually call tflint here.

Nuru: > The target from the original PR was renamed to `terraform/precommit`. You can call it here , or deprecate . I also noticed that you don't actually call tflint here.

We do call `terraform/precommit` here: https://github.com/cloudposse/build-harness/blob/6027a1d5567d8f45615dfcc5f562658d4db09fdf/templates/Makefile.build-harness#L148

which calls `terraform/tflint` here: https://github.com/cloudposse/build-harness/blob/6027a1d5567d8f45615dfcc5f562658d4db09fdf/modules/terraform/Makefile#L128

max-lobur (approved):

PR #350, opened 2023-05-24
Add tflint-fix target

max-lobur:
## what

Add target that fixes tflint some easy complaints

## why

Simplify maintenance

### DIFF
--- a/modules/terraform/Makefile
+++ b/modules/terraform/Makefile
@@ -34,6 +34,18 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+terraform/tflint-fix:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i "" 's/\([a-zA-Z{1,}][0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
+		sed -i "" 's/join("", \([^)]*\))/one(\1)/g' "$$v" ; \
+	done
+
+for v in $(find . -type f -name '*.tf'); do \
+		sed -i "" 's/\([a-zA-Z]+[0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$v" ; \
+		sed -i "" 's/join("", \([^)]*\))/one(\1)/g' "$v" ; \
+done
+
+
 ## Format check terraform
 terraform/lint:
 ifeq ($(OS), darwin)
@@ -125,5 +137,5 @@ terraform/v14-rewrite: terraform/loosen-constraints terraform/bump-tf-12-min-ver
 	@{ [[ "$(TERRAFORM_FORCE_README)" != "true" ]] && git diff --no-patch --exit-code README.yaml; } || $(MAKE) readme
 
 ## Terraform pull-request routine check/update
-terraform/precommit: terraform/fmt terraform/tflint readme/build
+terraform/precommit: terraform/tflint-fix terraform/fmt terraform/tflint readme/build
 	@# Need a recipe to avoid default recipe
--- a/modules/terraform/Makefile
+++ b/modules/terraform/Makefile
@@ -34,12 +34,12 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+# Auto-fix tflint errors
 terraform/tflint-fix:
 	@for v in $$(find . -type f -name '*.tf'); do \
-		sed -i 's/\([[:alpha:]]\+\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
+		sed -i 's/\([[:alpha:]]\+[0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
 		sed -i 's/join("", \([^)]*\))/one(\1)/g' "$$v" ; \
 	done
-
 
 ## Format check terraform
 terraform/lint:
--- a/modules/terraform/Makefile
+++ b/modules/terraform/Makefile
@@ -37,7 +37,7 @@ terraform/tflint:
 ## Auto-fix terraform deprecated list accessors
 terraform/tflint-fix-deprecated-index:
 	@for v in $$(find . -type f -name '*.tf'); do \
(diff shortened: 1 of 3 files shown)


modules/terraform/Makefile
```diff
@@ -34,6 +34,13 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+terraform/tflint-fix:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i 's/\([[:alpha:]]\+\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
```
max-lobur: This should be [[:alpha:]]\+[0-9]* to catch s3, ec2

modules/terraform/Makefile
```diff
@@ -125,5 +131,5 @@ terraform/v14-rewrite: terraform/loosen-constraints terraform/bump-tf-12-min-ver
 	@{ [[ "$(TERRAFORM_FORCE_README)" != "true" ]] && git diff --no-patch --exit-code README.yaml; } || $(MAKE) readme
 
 ## Terraform pull-request routine check/update
-terraform/precommit: terraform/fmt terraform/tflint readme/build
+terraform/precommit: terraform/tflint-fix terraform/fmt terraform/tflint readme/build
```
Nuru: Let's not run tflint fixes by default.

```suggestion
terraform/precommit: terraform/fmt terraform/tflint readme/build
```

modules/terraform/Makefile
```diff
@@ -34,6 +34,12 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+# Auto-fix tflint errors
+terraform/tflint-fix:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i 's/\([[:alpha:]]\+[0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
+	done
```
Nuru: Let's call this recipe something more specific, perhaps `tflint-fix-deprecated-index` and then have `tflint-fix` depend on it, both do document what this recipe fixes and to allow for future additions.

Also, I think the `sed` command should be

```
s/\([[:alpha:]]\+\)\.\([*0-9]\+\)\./\1[\2]\./g
```

but I don't think in practice it will matter.

Nuru (changes requested):

modules/terraform/Makefile
max-lobur: that sed wont catch ec2/s3

max-lobur:
@Nuru I addressed comments

modules/terraform/Makefile
```diff
@@ -34,6 +34,14 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
```
osterman, line 36: Please add a comment above this target explaining what it does.

modules/terraform/Makefile
```diff
@@ -34,6 +34,14 @@ terraform/validate:
...
 
+terraform/tflint-fix-deprecated-index:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i 's/\([[:alpha:]]\+[0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
+	done
+
+# Auto-fix tflint errors
```
osterman: Use `##` if we want this to show up in the `make help`, or single `#` for it to be hidden.

```suggestion
## Auto-fix tflint errors
```

modules/terraform/Makefile
max-lobur: done

max-lobur:
@osterman done

modules/terraform/Makefile
```diff
@@ -34,6 +34,15 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+## Auto-fix terraform deprecated list accessors
+terraform/tflint-fix-deprecated-index:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i 's/\([[:alpha:]]\+[0-9]*\)\.\([*0-9]\)\./\1[\2]\./g' "$$v" ; \
```
Nuru: I'm not happy with this regex. It would convert `v1.2.3` to `v1[2].3`. Let's use this to require at least 3 characters for the identifier part. I would rather miss some than change some incorrectly.

```suggestion
		sed -i 's/\([[:alpha:]][-_[:alpha:][:digit:]]\{2,\}\)\.\([*0-9]\+\)\./\1[\2]\./g' "$$v" ; \
```

modules/terraform/Makefile
```diff
@@ -34,6 +34,15 @@ terraform/validate:
...
+	done
+
```
Nuru, line 42:
```suggestion

.PHONY: terraform/tflint-fix terraform/tflint-fix-deprecated-index

```

Nuru (changes requested):

modules/terraform/Makefile
max-lobur: `ec2`, `s3` - won't work 😄  Tried that

max-lobur: I might try "and not semver" type of thing

modules/terraform/Makefile
```diff
@@ -34,6 +34,17 @@ terraform/validate:
 terraform/tflint:
 	tflint --enable-plugin=aws
 
+## Auto-fix terraform deprecated list accessors
+terraform/tflint-fix-deprecated-index:
+	@for v in $$(find . -type f -name '*.tf'); do \
+		sed -i 's/\([[:alpha:]][-_[:alpha:][:digit:]]\{2,\}\)\.\([*0-9]\+\)\./\1[\2]\./g' "$$v" ; \
```
Nuru, line 40: Sadly, this doesn't work on macOS. I think on macOS, character classes like `[:alpha:]` are not recognized in normal mode, and `-i` requires an argument. Not quite sure what to do about it.

Nuru (changes requested):

modules/terraform/Makefile
max-lobur: let's call only inside precommit/terraform which is inside the container?

max-lobur:
Not needed anymore, I've iterated through all repos with this anyways

PR #353, opened 2023-07-04
Highlight automated testing docs for terraform in a PR template

max-lobur:
## what

- Add automated testing docs for terraform

## why

- Keep reminding people the actual commands to use


Nuru:
Most PR's come from people who are not allowed to run automated testing. I think this is not worth the effort to update all the Terraform repos.

Nuru:
I am still opposed to this PR, but I reopened it so others can weigh in and have their say.

PR #360, opened 2023-10-24
Warning for make version

max-lobur:
## what
Warn on make version mismatch
It's just a warning message, intentionally not adding exit 1 to see how it goes. 

## why
We've had numerous issues with built-in macos make. Let's warn users if their version doesn't match what we expect

### DIFF
--- a/templates/Makefile.build-harness
+++ b/templates/Makefile.build-harness
@@ -67,6 +67,7 @@ endef
 .PHONY : init
 ## Init build-harness
 init::
+	if ! make --version | grep -q "GNU Make 4.4"; then echo "GNU Make 4.4 or higher is required. Please upgrade."; fi
 	@ $(harness_install)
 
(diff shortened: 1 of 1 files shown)


templates/Makefile.build-harness
```diff
@@ -67,6 +67,7 @@ endef
 .PHONY : init
 ## Init build-harness
 init::
+	if ! make --version | grep -q "GNU Make 4.4"; then echo "GNU Make 4.4 is required. Please upgrade."; fi
```
nitrocode: Have you considered adding some osx instructions on how to do this via homebrew?

```
brew install gmake
```

otherwise the user may not know how to upgrade `make`.

max-lobur: makes sense, adding

dudymas (approved):
approved!

PR #368, opened 2024-01-25
Use Organization's .github `README.md.gotmpl`

osterman:
## what
- Update `readme/build` to depend on template
- Fetch template if none exists

## why
- cloudposse/.github#28
- actions/checkout#1169
- actions/runner#2033


goruha (dismissed):

.github/workflows/lint.yml
```diff
@@ -14,7 +14,12 @@ jobs:
     name: readme
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v4
+      - name: Fix git safe.directory when run inside container
+        # https://github.com/actions/runner/issues/2033
+        run: |
+          mkdir -p /home/runner/work/_temp/_github_home
+          printf "[safe]\n\tdirectory = /github/workspace" > /home/runner/work/_temp/_github_home/.gitconfig
```
osterman: @Nuru should we just add this directly to the `Dockerfile`?

Nuru: We shouldn't be doing it this way at all.

Fixed.

Nuru (dismissed):

Nuru (approved):

.dockerignore
```diff
@@ -1 +1,2 @@
+.git/*
```
osterman: @Nuru these commands rely on 
`git remote get-url` which doesn't work with out `.git`, which means it won't work in `geodesic`

modules/readme/Makefile
```diff
@@ -8,27 +8,47 @@ export README_TEMPLATE_REPO_REMOTE ?= $(shell [ -d .git ] && git remote get-url
...
 
 export README_TEMPLATE_REPO ?= .github
-export README_TEMPLATE_REPO_REF ?= blob/main
+export README_TEMPLATE_REPO_REF ?= main
 export README_TEMPLATE_REPO_PATH ?= README.md.gotmpl
-export README_TEMPLATE_REPO_URL ?= https://github.com/$(README_TEMPLATE_REPO_ORG)/$(README_TEMPLATE_REPO)/$(README_TEMPLATE_REPO_REF)/$(README_TEMPLATE_REPO_PATH)?raw=true
+export README_TEMPLATE_REPO_URL ?= https://raw.githubusercontent.com/$${README_GITHUB_ORG}/$(README_TEMPLATE_REPO)/$(README_TEMPLATE_REPO_REF)/$(README_TEMPLATE_REPO_PATH)
 export README_TEMPLATE_FILE ?= $(BUILD_HARNESS_PATH)/templates/README.md.gotmpl
-export README_TEMPLATE_YAML := $(if $(findstring http,$(README_YAML)),$(README_YAML),$(BUILD_HARNESS_PATH)/templates/$(README_YAML))
-export README_INCLUDES ?= $(file://$(shell pwd)/?type=text/plain)
```
osterman, line 6: @Nuru this is needed by the `README.md.gotmpl` which uses it as the base for including other files (e.g. from `docs/`)

Nuru: With this syntax, `README_INCLUDES` is always empty. Probably meant for it to be automatically including `$PWD` but since it never worked, I would rather just remove it from the template.

- added in #100

PR #371, opened 2024-01-30
Restore README_INCLUDES

osterman:
## what
- Restore `README_INCLUDES`

## why
- #368 accidentally removed this, which is needed by the [`README.md.gotmpl`](https://github.com/cloudposse/.github/blob/main/README.md.gotmpl#L2)


Nuru:
Closing in favor of https://github.com/cloudposse/.github/pull/29

Note that the removal was not accidental, it was intentional, because the assignment was invalid. What my testing failed to discover was that the absence of the variable would cause issues that having the variable set but empty did not.

Issue #344, opened 2023-04-06
make readme fails to call installed gomplate binary


jonmchan:
### Describe the Bug

`make readme` installs gomplate to `vendor/gomplate`. However, when executing, https://github.com/cloudposse/build-harness/blob/master/modules/readme/Makefile#L32, the script tries to call the system version of gomplate instead of the installed version of gomplate.

For systems that do not have gomplate installed, the make readme fails. For example:

```
jonathanstimbp:bastion jonathan$ make readme
* Installing packages master...
Cloning into '/Users/jonathan/code/oss/bastion/build-harness/vendor/packages'...
remote: Enumerating objects: 921, done.
remote: Counting objects: 100% (921/921), done.
remote: Compressing objects: 100% (454/454), done.
remote: Total 921 (delta 316), reused 592 (delta 232), pack-reused 0
Receiving objects: 100% (921/921), 11.22 MiB | 26.42 MiB/s, done.
Resolving deltas: 100% (316/316), done.
* Installing gomplate to /Users/jonathan/code/oss/bastion/build-harness/vendor
mkdir -p -m a+rX /Users/jonathan/code/oss/bastion/build-harness/vendor
curl --retry 3 --retry-delay 5 --fail -sSL -o /Users/jonathan/code/oss/bastion/build-harness/vendor/gomplate https://github.com/hairyhenderson/gomplate/releases/download/v3.11.4/gomplate_darwin-amd64 && chmod +x /Users/jonathan/code/oss/bastion/build-harness/vendor/gomplate
make: gomplate: No such file or directory
make: *** [readme/build] Error 1
```

### Expected Behavior

`make readme` should succeed, utilizing the vendor/gomplate version instead of the system version. Perhaps the PATH variables are not being properly set or the script needs to be calling the local version instead.

### Steps to Reproduce

Do not have gomplate installed on the machine. Run `make readme` on any project.

jonmchan:
Please note related issue: https://github.com/cloudposse/bastion/issues/72

max-lobur:
I confirmed with the team that it's a bug. We supposed to have vendor at the PATH, but it's broken somewhere. This should work outside geodesic.

Nuru:
This is trickier than it may seem, because the build-harness Docker image is Linux and most people are not running this on Linux machines, which means the host binary of `gomplate` is incompatible with the binary build-harness would use from within the Docker container.

[...]

PR #373, opened 2024-01-31
Specify gomplate path in readme/Makefile

joe-niland:
## what

- add vendored gomplate path in readme/Makefile
- I understand this might be a naive solution - so I'd appreciate knowing why it wouldn't work, if that's the case. On my machine the darwin_amd64 binary is retrieved when installing via `make readme/deps`.

## why

- make readme/build requires external gomplate dependency, otherwise it errors

## references

- #344


Nuru (changes requested):
@joe-niland Thank you for this PR. I don't think this is the approach we want to take, but we will take it from here. Meanwhile, if you just `rm -rf ./build-harness/vendor/` or `make clean` to get rid of the offending binaries, and they will be automatically re-installed for the current OS and architecture.

osterman:
@joe-niland since gomplate is only one tool solved by this PR, but not the only tool affected, @Nuru wants to organize the binaries by architecture. I think he'll fix in this in a separate PR.

joe-niland:
@osterman thanks for the explanation

@Nuru ok, sounds good, thanks!

Closing

PR #376, opened 2024-02-02
Fix make/readme when github_repo has quotes

osterman:
## what
- Use `tr` to delete quote characters

## why
- repos will never have special shell characters in their name
- it breaks readme generation due to spurious quotes

## references
- #374 (introduced bug)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@
 build-harness.iml
 build-harness
 templates/README.md.gotmpl
-node_modules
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@
 build-harness.iml
 build-harness
 templates/README.md.gotmpl
+node_modules
```
osterman:
```suggestion
```

Nuru (approved):

PR #379, opened 2024-02-09
Drop phony requirement on `README_TEMPLATE_FILE`

osterman:
## what
- Drop the `.PHONY` for `README_TEMPLATE_FILE`

## why
- This breaks our usage of `make readme` in lots of other places where we generate READMEs based on local templates that differ from the org template. Here are [2 examples](https://github.com/search?q=repo%3Acloudposse%2Fdocs+make+readme+language%3APython&type=code&l=Python). 
- Worst case this file is cached locally. 

## references
- #368


Nuru (changes requested):
1. The breakage cited is not due to local templates being overwritten, it is due to parsing README.md rather than README.yaml
2. We have no other mechanism for ensuring that the README template gets updated when changed at the source

I am not saying we do not need to change anything, only that this is not the way to fix anything. I will propose a solution once I better understand the full extent of the problem.

osterman:
Templates are not proberly ending in `.gotmpl` or similar extension, but are templates.

https://github.com/cloudposse/docs/tree/master/scripts/docs-collator/templates

They are rendered here:
https://github.com/search?q=repo%3Acloudposse%2Fdocs+make+readme+language%3APython&type=code&l=Python

Which calls:
https://github.com/cloudposse/docs/blob/af73ab15ea1a9a4aec7d588772af866980f5a31f/scripts/docs-collator/ModuleRenderer.py#L63

PR #381, opened 2024-02-22
Also build&push arm docker image

hans-d:
## what

- also build & push arm version of docker images
- requires switching from alpine to debian (bookworm-slim)

## why

- running a platform native version on mac with arm cpu


hans-d:
seeing some differences with a local build. will investigate

PR #383, opened 2024-03-04
Ignore lockfiles while building Terraform docs

Nuru:
## what

- Ignore lockfiles while building Terraform docs

## why

- By default, `terraform-docs` takes into account any version pinning present in the automatically generated `.terraform.lock.hcl`. However, Cloud Posse modules do not use the lock files for version pinning, which means that different people can get different results when creating the documentation, which causes problems. By ignoring the lock files, we make the documentation easier to reproduce.


hans-d:
as per https://terraform-docs.io/user-guide/configuration/settings/#lockfile

hans-d (approved):

hans-d:
@Nuru needs cloudposse/engineering, i'm not yet part of that elite team ...

osterman (approved):

PR #388, opened 2024-09-24
Allows overriding README_TEMPLATE_REPO_URL and README_ALLOWLIST_ORGS.

### LINKED ISSUE #387: Unable to override README_TEMPLATE_REPO_URL and README_ALLOWLIST_ORGS
### Describe the Bug

These variables are defined with `:=` rather than `?=` which means it's not possible to override these variables, they will always be overwritten by the values set in the build-harness module Makfile. We use GitLab and need to override these variables to be able to load the template from GitLab and to be able to allowlist our organisation. Without these changes, it is not possible for us to use the README module without forking build-harness to add our changes.

### Expected Behavior

The variables do not override values set previously.

### Steps to Reproduce

Add to the Makefile before including the build-harness bootstrap:

```
export README_TEMPLATE_REPO_URL := https://gitlab.com/$(README_TEMPLATE_REPO_ORG)/$(README_TEMPLATE_REPO)/-/raw/$(README_TEMPLATE_REPO_REF)/$(README_TEMPLATE_REPO_PATH)
```

Notice how using the README module, this is overwritten to the original GitHub version.

### Screenshots

_No response_

### Environment

_No response_

### Additional Context

_No response_

irl:
## what

- Allows overriding README_TEMPLATE_REPO_URL and README_ALLOWLIST_ORGS.

## why

- See #387
- tl;dr: unable to use this for non-CloudPosse organisations, and unable to use with GitLab

## references

- Closes #387

### DIFF
--- a/modules/readme/Makefile
+++ b/modules/readme/Makefile
@@ -6,13 +6,13 @@ export README_TEMPLATE_REPO_ORG ?= $(shell [ -f "$(README_YAML)" ] &&  dirname $
 export README_TEMPLATE_REPO ?= .github
 export README_TEMPLATE_REPO_REF ?= main
 export README_TEMPLATE_REPO_PATH ?= README.md.gotmpl
-export README_TEMPLATE_REPO_URL := https://raw.githubusercontent.com/$(README_TEMPLATE_REPO_ORG)/$(README_TEMPLATE_REPO)/$(README_TEMPLATE_REPO_REF)/$(README_TEMPLATE_REPO_PATH)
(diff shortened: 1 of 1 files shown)


goruha:
@osterman @Nuru could you pls review this PR?
Is it fine with security?

RoseSecurity (commented):
After thinking about this and reviewing the code, I see a few potential areas this could be bad

Nuru:
I am vetoing this PR as a security risk. Since it is a security issue, I do not want to go into detail.

If you want to use a Readme template not provided by Cloud Posse, you can install it however you want and then set `README_TEMPLATE_FILE` to point to it.

